PR #7, opened 2019-02-14
Improve VFS simple's discovery of registered backends

Roosh513:
VFS simple currently assumes that you only want one default backend for each implementation (s3/gcs/os). This change will allow you to register backends based on URIs. It starts by searching for a backend based on a file URI, then falls back to bucket/volume, and finally falls back to the URI scheme.

### DIFF
--- a/vfssimple/doc.go
+++ b/vfssimple/doc.go
@@ -40,6 +40,41 @@ vfssimple is largely an example of how to initialize a set of backend filesystem
 initialization of the individual file systems.  See backend docs for specific authentication info for each backend but
 generally speaking, most backends can use Environment variables to set credentials or client options.
 
+File systems can only use one set of options. If you would like to configure more than one file system of the same type/schema with separate credentials,
+you can register and map file system options to locations or individual objects. The vfssimple library will automatically try to
+resolve the provided URI in NewFile() or NewLocation() to the registered file system.
+
+  package main
+
+  import(
+	"github.com/c2fo/vfs/vfssimple"
+	"github.com/c2fo/vfs/backend/s3"
+  )
+
+  ...
+
+  func DoSomething() error {
+	bucketAuth := s3.NewFileSystem().WithOptions(s3.Options{
+		AccessKeyID:     "key1",
+		SecretAccessKey: "secret1,
+		Region:          "us-west-2",
+	})
+
+	fileAuth := s3.NewFileSystem().WithOptions(s3.Options{
+		AccessKeyID:     "key2",
+		SecretAccessKey: "secret2,
+		Region:          "us-west-2",
+	})
+
+	backend.Register("s3://bucket1/, bucket1)
+	backend.Register("s3://bucket2/file.txt, fileAuth)
+
+    secureFile, _ := vfssimple.NewFile("s3://bucket2/file.txt")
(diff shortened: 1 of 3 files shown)


vfssimple/doc.go
```diff
@@ -40,6 +40,41 @@ vfssimple is largely an example of how to initialize a set of backend filesystem
...
+		Region:          "us-west-2",
+	})
+
+	fileAuth := s3.NewFileSystem().WithOptions(s3.Options{
+		AccessKeyID:     "key2",
+		SecretAccessKey: "secret2,
+		Region:          "us-west-2",
+	})
+
+	backend.Register("s3://bucket1/, bucket1)
```
funkyshu: the fs should be bucketAuth, not bucket1

vfssimple/doc.go
```diff
@@ -40,6 +40,41 @@ vfssimple is largely an example of how to initialize a set of backend filesystem
...
+	backend.Register("s3://bucket2/file.txt, fileAuth)
+
+    secureFile, _ := vfssimple.NewFile("s3://bucket2/file.txt")
```
funkyshu: indention is off

vfssimple/doc.go
```diff
@@ -40,6 +40,41 @@ vfssimple is largely an example of how to initialize a set of backend filesystem
...
+	publicLocation, _ := vfssimple.NewLocation("s3://bucket1/")
+
+	secureFile.CopyToLocation(publicLocation)
+  }
+
 To do more, especially if you need to pass in specific vfs.Option's via WithOption() or perhaps a mock client for testing via
 WithClient() or something else, you'd need to implement your own factory.  See github.com/c2fo/vfs/backend for more information.
```
funkyshu: Not sure this statement holds water anymore

vfssimple/doc.go
```diff
@@ -40,6 +40,41 @@ vfssimple is largely an example of how to initialize a set of backend filesystem
...
+File systems can only use one set of options. If you would like to configure more than one file system of the same type/schema with separate credentials,
+you can register and map file system options to locations or individual objects. The vfssimple library will automatically try to
+resolve the provided URI in NewFile() or NewLocation() to the registered file system.
+
+  package main
+
+  import(
+	"github.com/c2fo/vfs/vfssimple"
+	"github.com/c2fo/vfs/backend/s3"
+  )
```
funkyshu: will need "github.com/c2fo/vfs/backend" imported as well in this case.

funkyshu:
Don't forget to update the counterpart to vfssimple/doc.go, doc/vfssimple.md.

vfssimple/doc.go
Roosh513: I think it still does, since the whole purpose of using the backends packages directly is for building your own custom factory.

jkc: John is kinda right. You no longer necessarily have to create your own factory (as is being implied here), because you're making vfssimple more flexible/slightly less simple.

Roosh513: I read this sentence as a statement to do more beyond vfssimple. I'm open to wording it another way if anyone has suggestions.

funkyshu: This should be removed since you removed it also from the markdown version.

PR #18, opened 2019-05-17
Add ability to set a retrier for remote file systems

Roosh513:
This change covers a new interface which allows implementers to define a generic retrier in the event that the underlying file system client does not provide a retrier interface of their own. Retriers can be assigned via the filesystem options.

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -154,3 +154,8 @@ type Options interface{}
 //    return ret
 // }
 type Retry func(wrapped func() error) error
+
+// DefaultRetryer returns a no-op retryer which simply calls the wrapped command without looping.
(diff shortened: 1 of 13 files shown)


vfs.go
```diff
@@ -29,6 +29,9 @@ type FileSystem interface {
 
 	// Scheme, related to Name, is the uri scheme used by the FileSystem: s3, file, gs, etc...
 	Scheme() string
+
+	// Retry will return the retry function to be used by any file system.
+	Retry() Retry
```
Roosh513: Because I added this new function to the FileSystem interface, it will require us to do a full version bump of the library.

PR #20, opened 2019-06-06
in_mem_impl first PR

moezeid:
As of now, this commit implements (for File) these functions along with tests for each.
-Write()
-Read()
-Exists()
-Delete()
-LastModified()
-Size()
-Name()
-CopyToFile()
There is another branch being worked on that deals with location and its functions and puts it all together.

### DIFF
--- a/backend/mem/file.go
+++ b/backend/mem/file.go
@@ -126,7 +126,7 @@ func (File) CopyToLocation(location vfs.Location) (vfs.File, error) {
 func (f *File) CopyToFile(target vfs.File) error {
 	//if target exists, its contents will be overwritten, otherwise it will be created...i'm assuming it exists
 	_, err := target.Write(f.privSlice)
-	target.Close()
+	_ =target.Close()
 	return err
 
 }
--- a/backend/mem/file.go
+++ b/backend/mem/file.go
@@ -1,232 +1,188 @@
-package os
+package mem
 
 import (
-	"fmt"
-	"os"
-	"path"
-	"path/filepath"
-	"time"
-
+	"bytes"
+	"errors"
 	"github.com/c2fo/vfs/v4"
-	"github.com/c2fo/vfs/v4/utils"
+	"io"
+	"time"
 )
 
 //File implements vfs.File interface for os fs.
 type File struct {
-	file     *os.File
-	name     string
+	exists 		bool
+	timeStamp 	time.Time
+	isOpen		bool
+	isZB		bool
+	isRef		bool
+	privSlice 	[] byte
+	byteBuf  *bytes.Buffer
+	Filename string
+	cursor 		int
 	location vfs.Location
 }
 
-// newFile initializer returns a pointer to File.
-func newFile(name string) (*File, error) {
-	fileName := filepath.Base(name)
-	fullPath, err := filepath.Abs(name)
-	if err != nil {
-		return nil, err
-	}
-
-	fullPath = filepath.Dir(fullPath)
-
-	fullPath = utils.AddTrailingSlash(fullPath) //because "Dir" removes the trailing slash
-
-	location := Location{fileSystem: &FileSystem{}, name: fullPath}
-	return &File{name: fileName, location: &location}, nil
+func DoesNotExist() error {
+	return errors.New("This file does not exist!")
+}
+func CopyFail() error {
+	return errors.New("This file was not successfully copied")
 }
 
-// Delete unlinks the file returning any error or nil.
-func (f *File) Delete() error {
-	err := os.Remove(f.Path())
-	if err == nil {
-		f.file = nil
-	}
-	return err
-}
 
-// LastModified returns the timestamp of the file's mtime or error, if any.
-func (f *File) LastModified() (*time.Time, error) {
-	stats, err := os.Stat(f.Path())
-	if err != nil {
-		return nil, err
-	}
 
-	statsTime := stats.ModTime()
-	return &statsTime, err
-}
-
-// Name returns the full name of the File relative to Location.Name().
-func (f *File) Name() string {
-	return f.name
-}
-
-// Path returns the the path of the File relative to Location.Name().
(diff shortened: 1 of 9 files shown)


backend/mem/file.go
```diff
@@ -0,0 +1,219 @@
...
+	isZB		bool
+	isRef		bool
+	privSlice 	[] byte
+	byteBuf  *bytes.Buffer
+	Filename string
+	cursor 		int
+	location vfs.Location
+}
+
+func DoesNotExist() error {
```
jkc: Is there a reason for this and the following two functions to be public? I'm not against it, I just don't see the use case offhand. If there is a use case, this should be described in the function comments

backend/mem/file.go
```diff
@@ -0,0 +1,219 @@
...
+func (f *File) Close() error {  //NOT DONE
+	if !f.isRef {
+		// Do nothing on files that were never referenced
+		return nil
+	}
+	if f.byteBuf.Len() > 0{
+		f.privSlice = append(f.privSlice,f.byteBuf.Next(200)...) //TODO: maybe change the number for "Next" arg
+	}
+
+	f.isOpen =false
```
jkc: There are some spacing issues in this file that would (I think) be fixed by running go-fmt

backend/mem/file.go
```diff
@@ -0,0 +1,219 @@
...
+	"io"
+	"time"
+)
+
+//File implements vfs.File interface for os fs.
+type File struct {
+	exists 		bool
+	timeStamp 	time.Time
+	isOpen		bool
+	isZB		bool
```
jkc: This might be dumb, but what is "isZB"?

jkc: isRef and privSlice could probably use end of line comments describing their meaning/purpose as well...

backend/mem/file.go
```diff
@@ -0,0 +1,219 @@
...
+	return num, err
+}
+
+func (File) String() string {
+	panic("implement me")
+}
+
+func (f *File) Exists() (bool, error) {
+	if !f.exists {
+		return false, DoesNotExist()
```
jkc: I wouldn't necessarily expect Exists() to return an error for correctly returning that the file doesn't exist.

backend/mem/file.go
```diff
@@ -0,0 +1,219 @@
...
+	if existence {
+		return &f.timeStamp, err
+	}
+	return nil, err
+}
+
+func (f *File) Size() (uint64, error) {
+
+
+	return uint64(len(f.privSlice)),nil
```
jkc: remove extra lines above and below this.

backend/mem/location_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (s *osLocationTest) SetupSuite() {
+	setupTestFiles()
+}
+
+func (s *osLocationTest) TearDownSuite() {
+	teardownTestFiles()
+}
+/*
+func (s *osLocationTest) SetupTest() {
```
jkc: Will this be uncommented at some point, or should it be removed entirely?

backend/mem/location_test.go
```diff
@@ -0,0 +1,186 @@
...
+	fileLocation := otherFile.Location()
+	subDir, _ := fileLocation.NewLocation("other/")
+	s.Equal("/foo/other/", subDir.Path())
+
+	relDir, _ := subDir.NewLocation("../../bar/")
+	s.Equal("/bar/", relDir.Path(), "relative dot path works")
+}
+
+func (s *osLocationTest) TestNewFile() {
+	//loc, err := s.fileSystem.NewLocation("", "/foo/bar/baz/")
```
jkc: Same comment as above here, should this test be removed, or will it be fixed in a later commit?

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,49 @@
+package mem
+
+import (
+	"github.com/c2fo/vfs/v4"
+	"github.com/c2fo/vfs/v4/backend"
+	"github.com/c2fo/vfs/v4/utils"
+)
+
+//Scheme defines the filesystem type.
+const Scheme = "file"
```
funkyshu: "file" is the scheme used for os files.
I suggest that Scheme should be "mem" or "memfs", such that when building a URI, like:
```
         foo://example.com:8042/over/there
         \_/   \______________/\_________/
          |           |            |
       scheme       volume        path
```
we would end up with:
`mem://myNamespace/some/path/to/file/`

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,49 @@
...
+const name = "mem"
```
funkyshu: name is should be a textual name of the filesystem type.
I'd suggest something like "In-Memory Filesystem"

backend/s3/options_test.go
```diff
@@ -47,3 +49,74 @@ func (o *optionsTestSuite) TestGetClient() {
...
+	panic("implement me")
+}
+
+func (*Foo) Name() string {
+	panic("implement me")
+}
+
+func (*Foo) URI() string {
+	panic("implement me")
+}
```
funkyshu: Not sure why any changes are being made to thsi file.

backend/mem/file.go
thecsw: @jkc `gofmt` doesn't fix the problems with empty lines. It will remove the whitespace in an empty line, like `s/\s*$//g` but not the problems with empty lines after function signatures.

I wrote this sed script that fixes them: 

``` sh
sed -e '1h;2,$H;$!d;g' -e 's|{\n\s*|{|g;s|\s*\n}|}|g'
```

jkc: I was in part referring to the extra spacing throughout the file, but in this particular comment I was mostly pointing to the lack of a space before 'false' ie: '=false' should be '= false'

moezeid:
new PR has an updated and improved version of this

PR #22, opened 2019-06-13
Location_in_mem and all that preceded (corrected)

moezeid:
Completed what was missing from previous PR.  Everything is implemented and tests pass.  Let me know!


backend/mem/fileSystem.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/c2fo/vfs/v4/utils"
+	"path"
+)
+
+//Scheme defines the filesystem type.
+const Scheme = "mem"
+const name = "In-Memory Filesystem"
+
+var systemMap map[string]*File
+var fileList []*File
```
funkyshu: Rather than making these global, can we make them a part of the FileSystem struct?

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,78 @@
...
+func (fs *FileSystem) Retry() vfs.Retry {
+	return vfs.DefaultRetryer()
+}
+
+// NewFile function returns the mem implementation of vfs.File.  NOT DONE
+func (fs *FileSystem) NewFile(volume string, name string) (vfs.File, error) {
+
+	if !path.IsAbs(name) {
+		name = path.Join("/", name)
+	}
```
funkyshu: We cant just assume they meant to have a leading slash.  In fact that might be really dangerous.
If a user somehow expected relative path, and they put 
```
file, _ := fs.Newfile("", "etc/passwd")
file.Delete()
```
it would delete /etc/passwd.

Instead we should be doing validation and return an error if necessary.

backend/mem/file.go
```diff
@@ -0,0 +1,469 @@
...
+func newFile(name string) (*File, error) {
+
+	//var l Location
+	//tmp, err := (*Location).NewFile(&l,name)
+	//file := tmp.(*File)
+
+	return &File{
+		timeStamp: time.Now(), isRef: false, Filename: path.Base(name), cursor: 0,
+		isOpen: false, exists: false,
+	}, nil
```
funkyshu: if you only record the base path, how do we know what location we're at?

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,78 @@
...
+
+	file, _ := newFile(name)
+	tmp, err := fs.NewLocation(volume, name)
+	file.location = tmp
+	if systemMap[name] != nil && systemMap[name].getIndex() != -1 {
```
funkyshu: This works in javascript but not in go.
What you want is
```
if _, ok := systemMap[name]; ok {
    err = systemMap[name].Delete()
    if err != nil {
        return nil, err
    }
}
```

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,78 @@
...
+
+// NewFile function returns the mem implementation of vfs.File.  NOT DONE
+func (fs *FileSystem) NewFile(volume string, name string) (vfs.File, error) {
+
+	if !path.IsAbs(name) {
+		name = path.Join("/", name)
+	}
+
+	file, _ := newFile(name)
+	tmp, err := fs.NewLocation(volume, name)
```
funkyshu: Here you're using name... if name was /some/path/to/file.txt you're passing that to the NewLocation() when all you should be passing is path.Dir(name)

backend/mem/file.go
```diff
@@ -0,0 +1,469 @@
...
+		f.cursor++
+		if i >= length || i >= len(f.privSlice) {
+			break
+		}
+		if i == length {
+			break
+		}
+		p[i] = f.privSlice[i]
+	}
+	f.timeStamp = time.Now()
```
funkyshu: why are we modifying the time on a read?  If it was last accessed, that would make sense but we only care about last modified.

backend/mem/file.go
```diff
@@ -0,0 +1,469 @@
...
+}
+
+func moveToLocationError() error {
+	return errors.New("Move to location unexpectedly failed")
+}
+func writeError() error {
+	return errors.New("Unexpected Write Error")
+}
+
+func seekError() error {
```
Roosh513: The way these are used is actually hiding the underlying error. You are catching the error and returning this error in your functions without wrapping the original error. It's probably better just to return the error directly in most cases.

backend/mem/file.go
```diff
@@ -0,0 +1,469 @@
...
+		return false, nil
+	} else {
+		return true, nil
+	}
+}
+
+//Location simply returns the file's underlying location struct pointer
+func (f *File) Location() vfs.Location {
+
+	newLoc, _ := f.location.NewLocation("")
```
funkyshu: This will be problematic because of the ChangeDir bug you identified.

funkyshu:
I haven't looked at Location yet... but figured you could get started with what I've had so far.

backend/mem/fileSystem.go
Roosh513: I'm going to play devil's advocate here and disagree. Working at the file system level should assume you are using the volume as your starting point, because it has no location context.

Just slightly modifying your example for additional context demonstrates this:

```
file, _ := fs.Newfile("C:", "etc/passwd")
file.Delete()
```

It should be up to the developer to do an "Exist()" check on the returned file before deleting. Some operating systems won't blink an eye or ask twice if you type:

`rm -rf /`

I don't think our file system implementations should behave any different. This is something we should probably clarify in the interface documentation.

jkc:
.DS_Store should be in your .gitignore file. It is an OSX file that shouldn't be committed to the repo

backend/mem/file.go
```diff
@@ -0,0 +1,469 @@
...
+	"time"
+)
+
+//File implements vfs.File interface for os fs.
+type File struct {
+	exists    bool
+	timeStamp time.Time
+	isOpen    bool
+	isRef     bool         //has it been referenced before?
+	privSlice []byte       //the file contents
```
jkc: Rather than naming this after the data structure ie: what it means to the language (it is literally a private slice) name this (and other variables) for what it means to us (the developers that will be writing and maintaining the code). According to the comment maybe a more appropriate name for this variable is 'contents'? 'File.contents' is a lot more meaningful than 'File.privSlice'. 

Similarly above, if someone sees 'file.isRef' somewhere, will they wonder whether this means a boolean to refer to whether the file *is* a reference vs whether it has been referenced? Does 'isReferenced' make more sense? I'm literally asking, not trying to lead you to my line of thinking, I don't actually know yet how 'isRef' is used.

And 'timeStamp'... yep, it is a time.Time. What does this timeStamp mean to the context of the file from our point of view when we see this property referenced elsewhere in the code? Is it a created timestamp? modified/last accessed timestamp? Any touch to the file (all of the above?) in which case would something like file.lastTouched, or file.lastModified be more descriptive?

I know I asked you to comment these properties in my last review, I should have thought about what I was requesting a little more. If you have to comment property/member names to describe what they are, it is likely a sign that the name needs to be rethought. The comments will help when someone is looking at this struct definition, but a good name will be much more useful throughout the code.

backend/mem/location.go
```diff
@@ -0,0 +1,211 @@
...
+List finds all of the files living at the current location and returns them in a slice of strings.
+If there are no files at location, then an empty slice will be returned
+*/
+func (l *Location) List() ([]string, error) {
+	//panic("implement me")
+
+	list := make([]string, 0)
+	str := l.Path()
+	for _, v := range fileList {
+		if v != nil {
```
thecsw: A bit of nitpicking here, but to reduce the number of indentation (which is the True Go Style), you can do 

``` go
for _, v := range fileList {
  if v == nil {
   continue
  }
  // my stuff but without an additional indent.
```

backend/mem/location.go
```diff
@@ -0,0 +1,211 @@
...
+					existence, _ := systemMap[fullPath].Exists()
+					if existence {
+						list = append(list, v.Name())
+					}
+				}
+			}
+		}
+	}
+	return list, nil
+}
```
thecsw: Lots of nested if loops here. You can always do inverse boolean logic to espace the event loop. It would be cleaner. Or if statements are logically compatible, you can make a chained boolean expression with `&&`s

backend/mem/location.go
```diff
@@ -0,0 +1,211 @@
...
+ListByPrefix tags a prefix onto the current path and in a slice,
+ returns all file base names whose full paths contain that substring
+Returns empty slice if nothing found
+*/
+func (l *Location) ListByPrefix(prefix string) ([]string, error) {
+
+	list := make([]string, 0)
+	str := path.Join(l.Path(), prefix)
+	for _, v := range fileList {
+		if v != nil {
```
thecsw: Same here, you can do an inverse boolean logic trick here.

backend/mem/location.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+//Volume returns the volume of the current fs. In-Memory-Fs has no volume
+func (l *Location) Volume() string {
+	return ""
+}
+
+//Path returns the full, absolute path of the location with leading and trailing slashes
+func (l *Location) Path() string {
+
```
thecsw: Weird spaces here. Doesn't `gofmt`, `goimports`, or `go-vet` fix this?

backend/mem/location.go
```diff
@@ -0,0 +1,211 @@
...
+	l.name = path.Join(l.name, relativePath)
+	return nil
+
+}
+
+//FileSystem returns the type of filesystem location exists on, if it exists at all
+func (l *Location) FileSystem() vfs.FileSystem {
+
+	filePath := path.Join(path.Clean(l.Path()), l.Filename)
+	if systemMap[filePath] != nil {
```
thecsw: That's not how you check maps. Go maps also return a second variable whose boolean value will tell you if the value exists in the map.

``` go
if _, ok := mymap[badvalue]; ok {
    //my code
}
```

backend/mem/location.go
```diff
@@ -0,0 +1,211 @@
...
+	systemMap[nameStr] = file
+	fileList = append(fileList, file)
+
+	return file, nil
+
+}
+
+//DeleteFile locates the file given the fileName and calls delete on it
+func (l *Location) DeleteFile(fileName string) error {
+
```
thecsw: Go's style is against whitespacing after function's signature.

Run this shell script and it will fix it everywhere (took me 10 minutes to write lol)

``` sh
sed -e '1h;2,$H;$!d;g' -e 's|{\n\s*|{|g;s|\s*\n}|}|g'
```

If you have sane GNU coreutils, run `find . -type f | grep '\.go$'` | xargs sed -e '1h;2,$H;$!d;g' -e 's|{\n\s*|{|g;s|\s*\n}|}|g' -i` and it will fix all of that formatting!

backend/mem/location_test.go
```diff
@@ -0,0 +1,233 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/c2fo/vfs/v4"
+)
+
+/**********************************
+ ************TESTS*****************
```
thecsw: We know that, `_test.go` files are *tests* only.

thecsw:
Why is this PR practically a copy of the first one? https://github.com/C2FO/vfs/pull/20

You can close the first PR because this is newer.

backend/mem/location.go
Roosh513: I would avoid running this command against the entire project. Instead, ensure that your IDE is configured to automatically handle formatting.

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,78 @@
...
+		fileSystem: fs,
+		name:       utils.AddTrailingSlash(path.Clean(name)),
+		exists:     false,
+	}, nil
+
+}
+
+// Name returns "mem"
+func (fs *FileSystem) Name() string {
+	return name
```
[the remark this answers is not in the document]
jkc: It is part of the interface, and must be implemented to meet the requirements, even if it isn't anything more than an accessor to an underlying property. This guarantees that anyone implementing a FileSystem will grant access to that file system's name, which we wouldn't necessarily be able to enforce by just asking them to make a public property.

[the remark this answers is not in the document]
thecsw: @jkc Thanks for explaining it! My bad that I didn't look into interfaces

PR #23, opened 2019-06-18
Ep 2432

### LINKED ISSUE #10: GCS location ListByPrefix() not working properly with vfssimple locations
When getting a location via a GCS URI through vfssimple.NewLocation(), I was unable to list functions via the location instance's ListByPrefix() function. When digging through the source, it looks like this function uses a leading slash ('/') character in the query to GCS. This function should ensure that there is no leading slash on the prefix defined on the location struct.

funkyshu:
### Changed
- Defined stricter definitions in interfaces, largely related to input requiements being more
URI compliant.
- FileSystem's NewFile and NewLocation required absolute paths.  Locations NewFile and NewLocation require relative paths.
- Added/Updated several now utils functions for URI validation across implementations.
- Ensure native methods are being used for atomic functions like copy or move/rename where possible
- Document expected behavior with empty readers in all copy and move function as well as io.Copy
- ListByPrefix now allows relative paths.
### Added
- Integration test suite for backened developers that ensures expected interface behaviors and tests interactions
between implementations.
- Examples of io.Copy usage in docs.
### Fixed
- Fixes #10  Fixed vfssimple bug that would return os filesystem if the uri had the word 'file' anywhere in it even if was s3, like s3://bucket/path/to/file.txt
- Fixed bug where MoveToLocation was returning the old file with updated location rather
than a brand new file with new location.  Since File may represent a non-existant file, the original file should still be valid.


jkc:
Good god man.

backend/gs/file.go
```diff
@@ -200,8 +198,8 @@ func (f *File) MoveToLocation(location vfs.Location) (vfs.File, error) {
 // MoveToFile puts the contents of File into the targetFile passed using File.CopyToFile.
```
jkc: not sure the comments need to be directly referencing variable names in the first place, but you changed the variable/arg name, and it no longer matches the one referenced in this comment. This should probably just say "...into the file argument passed..." or something similar.

backend/gs/location.go
```diff
@@ -108,18 +110,33 @@ func (l *Location) Exists() (bool, error) {
...
 // NewLocation creates a new location instance relative to the current location's path.
 func (l *Location) NewLocation(relativePath string) (vfs.Location, error) {
-	newLocation := *l
+	if l == nil {
+		return nil, errors.New("non-nil gs.Location pointer is required")
+	}
+
+	//make a copy of the original location first, then ChangeDir, leaving the original location as-is
+	newLocation := &Location{}
+	*newLocation = *l
```
jkc: As far as I know this is fine, but I definitely don't understand what exactly is happening here. You'll need to go over this with me tomorrow, when I'm in the office.

utils/utils.go
```diff
@@ -4,44 +4,69 @@ import (
...
-	slash := "/"
-	if runtime.GOOS == Windows {
-		slash = "\\"
+// RemoveLeadingSlash removes leading slash, if any
+func RemoveLeadingSlash(path string) string {
+	return strings.TrimLeft(path, "/")
+}
+
+// ValidateAbsFilePath ensures that a file path has a leading slash but not a trailing slash
+func ValidateAbsFilePath(name string) error {
```
jkc: Is there a reason "Rel" and "Abs" are abbreviated, when as a general rule all other parts of other function names are spelled out?

backend/gs/location.go
funkyshu: This makes this dereferences the pointer, makes a copy, then returns a pointer to the new copy (and not the original struct).

PR #24, opened 2019-06-21
In-Memory Backend 

moezeid:


backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+	}, nil
+
+}
+
+// Name returns "mem"
+func (fs *FileSystem) Name() string {
+	return name
+}
+
+// Scheme return "file" as the initial part of a file URI ie: file://
```
funkyshu: this is not a os backend.

backend/mem/doc.go
```diff
@@ -0,0 +1,30 @@
+/*
+Package os built-in os lib VFS implementation.
```
funkyshu: This is no an os backend

backend/mem/test_files/foo.txt
```diff
@@ -0,0 +1 @@
+Hello World!
```
funkyshu: I can't see anywhere where this file is used.  Even if you did have an os File, it would be preferable you set it up as needed as part of your test (and removed it after) in a temp are (ie, using ioutil.TempFile(), or ioutil.TempDir())

backend/mem/file.go
```diff
@@ -0,0 +1,497 @@
...
+	if copyErr != nil {
+		return copyErr
+	}
+
+	return f.Delete()
+
+}
+
+/*
+Delete removes the file from the fs. Sets it path in the fsMap to nil,
```
jkc: fs should be file system, spell it out unless you're referring to variables etc... (as in 'fsMap' in this line). Here you're referring to removal from the underlying file system, not an fs struct.

backend/mem/file.go
```diff
@@ -0,0 +1,497 @@
...
+ and also nils the file's members
+*/
+func (f *File) Delete() error {
+
+	if ex, _ := f.Exists(); !ex {
+		return doesNotExist()
+	}
+
+	if f == nil {
```
jkc: Should this conditional come before the prior Exists check? ie: it is meaningless to take hardly any action on a nil file pointer, so this check should come before a possible nil.Exists(). I'm not 100% sure of this, if you have a reason let me know.

backend/mem/file.go
```diff
@@ -0,0 +1,497 @@
...
+			thisObj.i = nil
+			thisObj = nil
+			(*mapRef)[loc.Volume()][str] = nil //setting that key to nil so it truly no longer lives on this system
+		}
+	}
+
+	return nil
+}
+
+//newFile creates an in-mem vfs file given the name then returns it
```
jkc: "in-mem" should be "in-memory".

backend/mem/file.go
```diff
@@ -0,0 +1,497 @@
...
+//Path returns the absolute path to the file
+func (f *File) Path() string {
+
+	return path.Join(f.Location().Path(), f.name)
+
+}
+
+//Name returns the file's base name
+func (f *File) Name() string {
+	//if file exists
```
jkc: what does this comment refer to? it is returning f.name regardless of file existence. Remove, if it no longer applies.

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/c2fo/vfs/v5"
+	"github.com/c2fo/vfs/v5/backend"
+	"github.com/c2fo/vfs/v5/utils"
+)
+
+//Scheme defines the filesystem type.
+const Scheme = "mem"
+const name = "In-Memory Filesystem"
+
+// FileSystem implements vfs.Filesystem for the mem filesystem.
```
jkc: This comment should be directly above the FileSystem struct, not with the obj and objMap between, also, be sure to follow the camel casing of FileSystem (vfs.FileSystem, not vfs.Filesystem) and because of that, all places where file system is spelled out, it should be two words. (see the line above, and line 12 and 14 also needs to be fixed.)

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+type obj struct {
+	isFile bool
+	i      interface{}
+}
+type objMap map[string]*obj
+type FileSystem struct {
+	fsMap map[string]objMap
+}
+
+// FileSystem will return a retrier provided via options, or a no-op if none is provided.
```
jkc: Start this comment with "Retry will..." not "FileSystem will..." The comment should start with the name of the function being described.

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+func (fs *FileSystem) Retry() vfs.Retry {
+	return vfs.DefaultRetryer()
+}
+
+/*
+NewFile function returns the mem implementation of vfs.File.
```
jkc: memory, or in-memory

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+Since this is inside "fileSystem" we assume that the caller knows that the CWD is the root.
```
jkc: "fileSystem" should be capitalized if referring to the type, two words, if referring to the underlying concept.

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+If a non-absolute path is given, an error is thrown.Additionally, a file does not
```
jkc: Missing space in "thrown.Additionally"

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+technically exist until a call to "Touch()" is made on it. The "Touch" call links the
+file with the fileSystem's map and brings it into existence.
```
jkc: Capitalize reference to FileSystem type

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+	return &Location{
+		fileSystem: fs,
+		name:       str,
+		exists:     false,
+		volume:     volume,
+	}, nil
+
+}
+
+// Name returns "mem"
```
jkc: Something along the lines of:
// Name returns the name of the file system.
would be more accurate and prevent you from needing to update the comment if the name constant ever gets updated. IE: it doesn't actually return "mem" it returns "In-Memory Filesystem" as defined currently.

backend/mem/fileSystem.go
jkc: This hasn't been resolved...  the comment is still incorrectly referring to "file" and "file://" when it should be "mem" and "mem://"

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+func (fs *FileSystem) Name() string {
+	return name
+}
+
+// Scheme return "file" as the initial part of a file URI ie: file://
+func (fs *FileSystem) Scheme() string {
+	return Scheme
+}
+func (fs *FileSystem) Initialize() {
```
jkc: All public methods need a comment "// Initialize <does whatever it does...>" 

That said, is this Initialize function something that should ever be used by an end user?

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+//getKeys is used to get a list of absolute paths on a specified volume. These paths are a mixture of files and locations
+func (o objMap) getKeys() []string {
+	keyList := make([]string, 0)
+	for i := range o {
+		keyList = append(keyList, i)
+	}
+	return keyList
+}
+
+//fileHere returns a list of file pointers found at the absolute location path provided.  If none are there, returns an empty slice
```
jkc: comment missing an s "filesHere" is the name of the function, not "fileHere"

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,176 @@
...
+			file := object.i.(*File)
+			if file.Location().Path() == absLocPath {
+				fileList = append(fileList, file)
+			}
+		}
+	}
+	return fileList
+}
+
+//fileHere returns a list of file pointers found at the absolute location path provided.  If none are there, returns an empty slice
```
jkc: wrong function name in the comment... actually the whole thing is copy/past of filesHere comment... should be updated with whatever fileNamesHere does.

backend/mem/location_test.go
```diff
@@ -0,0 +1,301 @@
...
+ ************TESTS*****************
+ **********************************/
+
+type memLocationTest struct {
+	suite.Suite
+	testFile   vfs.File
+	fileSystem *FileSystem
+}
+
+func (s *memLocationTest) SetupSuite() {
```
jkc: Is this function necessary?

backend/mem/location_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+}
+
+func (s *memLocationTest) TearDownSuite() {
```
jkc: Same as above... This function can probably be removed.

backend/mem/location_test.go
```diff
@@ -0,0 +1,301 @@
...
+func (s *memLocationTest) SetupTest() {
+	fs := &FileSystem{}
+	fs.Initialize()
+
+	file, nerr := fs.NewFile("", "/test_files/test.txt")
+	s.NoError(nerr, "File creation was not successful so it does not exist")
+
+	s.testFile = file.(*File)
+	s.fileSystem = fs
+	//s.fileSystem.Initialize()
```
jkc: remove commented code.

backend/os/location.go
```diff
@@ -19,15 +19,13 @@ type Location struct {
 	fileSystem vfs.FileSystem
 }
 
-// NewFile uses the properties of the calling location to generate a vfs.File (backed by an os.File). A string
+// NewFile uses the properties of the calling location to generate a vfs.File (backed by a osFile). A string
```
jkc: Why was this changed?

moezeid: no idea. I must have changed it at some point and i couldn't remember whether the period should have been in there or not.

backend/mem/fileSystem.go
moezeid: my bad! fixed it in some other place and thought i fixed it here

backend/mem/fileSystem.go
moezeid: agreed on the comments. I think i would need it to be public since that is how the map is initialized.  if the end user would be the one created instances of this FileSystem, then I would say yes it needs to be public.

jkc: K. Then the only issue here is it should be commented. Thanks!

backend/mem/location.go
```diff
@@ -0,0 +1,202 @@
...
+ListByPrefix tags a prefix onto the current path and in a slice,
+ returns all file base names whose full paths contain that substring
+Returns empty slice if nothing found
+*/
+
+func (l *Location) ListByPrefix(prefix string) ([]string, error) {
+
+	list := make([]string, 0)
+	str := path.Join(l.Path(), prefix)
+	mapRef := &l.fileSystem.fsMap
```
Roosh513: I see this patter a lot:

```
mapRef := &l.fileSystem.fsMap
(*mapRef)[l.volume]
```

Why not just get the map instead of a pointer to the map? It cleans up syntax a lot.


```
mapRef := l.fileSystem.fsMap
mapRef[l.volume]
```

backend/mem/location.go
```diff
@@ -0,0 +1,202 @@
...
+	l.exists = true
+	return true, nil
+}
+
+/*
+NewLocation creates a new location at the
+given relative path, which is tagged onto the current locations absolute path
+*/
+func (l *Location) NewLocation(relLocPath string) (vfs.Location, error) {
+	err := utils.ValidateRelativeLocationPath(relLocPath)
```
Roosh513: This function should check the file system to see if a location already exist. Return early with that location's reference if it exists.

backend/mem/location.go
```diff
@@ -0,0 +1,202 @@
...
+//FileSystem returns the type of file system location exists on, if it exists at all
+func (l *Location) FileSystem() vfs.FileSystem {
+
+	return l.fileSystem
+
+}
+
+//NewFile creates a vfs.File given its relative path and tags it onto "l's" path
+func (l *Location) NewFile(relFilePath string) (vfs.File, error) {
+
```
Roosh513: Same here. This function should check the file system to see if a file already exists at the location with the given name. Return early with that file's reference if it exists.

```
obj, ok := l.fileSystem.fsMap[l.volume][path.Join(l.Path(), relFilePath)]
if ok && obj.isFile {
		return obj.i.(*File), nil
}
```

backend/mem/file_test.go
```diff
@@ -0,0 +1,723 @@
...
+
+//TestCopyToNilFile ensures that an error is thrown for trying to copy to a nil file
+func (s *memFileTest) TestCopyToNilFile() {
+
+	newFile, err := s.fileSystem.NewFile("", "/nilFile.txt")
+	s.NoError(err, "Unexpected error creating file")
+	newFile = nil
+	err = s.testFile.CopyToFile(newFile)
+	s.Error(err, "Expected error for copying to nil file")
+}
```
funkyshu: we can just remove lines 334 and 335.  They aren't doing anything in regards to our test.

backend/mem/location.go
```diff
@@ -0,0 +1,202 @@
...
+	"time"
+
+	"github.com/c2fo/vfs/v5"
+	"github.com/c2fo/vfs/v5/utils"
+)
+
+//Location implements the vfs.Location interface specific to in-memory FileSystem.
+type Location struct {
+	exists     bool
+	firstTime  bool
```
funkyshu: firstTime isn't used

backend/mem/file.go
```diff
@@ -0,0 +1,444 @@
...
+	f.cursor = 0
+	return nil
+}
+
+//Read implements the io.Reader interface.  Returns number of bytes read and potential errors
+func (f *File) Read(p []byte) (n int, err error) {
+
+	//if file exists:
+	offset := f.cursor
+	if f.isOpen == false {
```
funkyshu: this can be simplified to:
```
	if !f.isOpen {
```

backend/mem/file.go
```diff
@@ -0,0 +1,444 @@
...
+	default:
+		return 0, seekError()
+
+	}
+
+}
+
+//Write implements the io.Writer interface. Returns number of bytes written and any errors
+func (f *File) Write(p []byte) (n int, err error) {
+	if f.isOpen == false {
```
funkyshu: this can also be simplified to:
```
	if !f.isOpen {
```

backend/mem/fileSystem.go
```diff
@@ -0,0 +1,183 @@
...
+func (o objMap) getKeys() []string {
+	keyList := make([]string, 0)
+	for i := range o {
+		keyList = append(keyList, i)
+	}
+	return keyList
+}
+
+//fileHere returns a list of file pointers found at the absolute location path provided.  If none are there, returns an empty slice
+func (o objMap) filesHere(absLocPath string) []*File {
```
funkyshu: This function seems unused.

backend/mem/location.go
```diff
@@ -0,0 +1,202 @@
...
+
+	err := utils.ValidateRelativeFilePath(relFilePath)
+	if err != nil {
+		return nil, err
+	}
+	pref := l.Path()
+	str := relFilePath
+	var nameStr string
+
+	nameStr = path.Join(pref, str)
```
funkyshu: Simplify:
```
	var nameStr string

	nameStr = path.Join(pref, str)
```
to:
```
	nameStr := path.Join(pref, str)
```

backend/mem/file_test.go
```diff
@@ -0,0 +1,723 @@
...
+	s.Equal("mem:///test_files/lots/of/directories/here/we/go/test.txt", str)
+
+}
+
+func TestMemFile(t *testing.T) {
+	suite.Run(t, new(memFileTest))
+	_ = os.Remove("test_files/new.txt")
+}
+
+func (s *memFileTest) teardownTestFiles() {
```
funkyshu: this function is unused

backend/mem/file.go
```diff
@@ -0,0 +1,444 @@
...
+
+	//if file exists:
+	offset := f.cursor
+	if f.isOpen == false {
+		f.isOpen = true
+	}
+	existence, err := f.Exists()
+	if !existence {
+		return 0, err
+	}
```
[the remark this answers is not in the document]
jkc: Seems like this should be more along the lines of:
`exists, err := f.Exists() // using 'exists' vs 'existence' because it makes more sense grammatically in conditionals ie: 'if [it] exists' vs 'if [it] existence'
//handle Exists errors first
if err != nil {
    return 0, err
}
if !exists {
    return 0, nil // not sure offhand whether it would be more consistent with other implementations to error for read on non-existent file or not.
}`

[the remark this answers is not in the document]
jkc: well, that formatted badly.

backend/mem/file.go
```diff
@@ -0,0 +1,444 @@
...
+	readBufLen := len(p)
+	if readBufLen == 0 { //readBufLen of byte slice is zero, just return 0 and nil
+		return 0, nil
+	}
+	fileContentLength := len(f.contents)
```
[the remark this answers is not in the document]
jkc: Contrary to how I probably generally come off in these discussions I'm actually fine with using small, even single-letter variables (the Go Way) as long as the variable's use is limited to a 5 to maybe... 10-20 line function/block. But if you are going to spell out a variable, spell it out. Why shortcut 2-4 characters when not doing so guarantees legibility?

My main reasons for this is: 1. there is no ambiguity or personal opinion for what words/letters to abbreviate, and what will be obvious to everyone, and 2. it is, granted, probably obvious what readBufLen means when we look at it, but if I'm adding to or refactoring this later, and I'm doing something with your 'readBufLen' variable, I have to remember how you abbreviated it... is it 'readBufferLength'? 'readBuffLen'? 'readBufLength'? etc... better to spell it out entirely, or if the scope is limited enough, abbreviate it absolutely: 'l' In this case, you're working with multiple 'lengths' so just 'l' might be a bad example...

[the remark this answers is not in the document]
jkc: Sorry, this may not be the best place to have the ongoing variable naming conversation... mostly I agree with Judd's comment, as far as this specific request goes...

PR #25, opened 2019-07-08
Touch

funkyshu:


backend/gs/file.go
```diff
@@ -220,6 +220,132 @@ func (f *File) Delete() error {
...
+		return f.updateLastModifiedByMoviing()
+	}
+
+	return f.updateLastModifiedByAttrUpdate()
+}
+
+func (f *File) updateLastModifiedByMoviing() error {
+	// setup a tempfile
+	tempfile, err := f.Location().
+		NewFile(fmt.Sprintf("%s.%d", f.Name(), time.Now().UnixNano()/1000000))
```
Roosh513: There's probably a time constant that you can use for the denominator here.

Roosh513:
Is there a reason to keep TouchCopy around other than passivity?

funkyshu:
> Is there a reason to keep TouchCopy around other than passivity?

Mostly because of the note described at https://github.com/C2FO/vfs/pull/25/files#diff-04c6e90faac2675aa89e2176d2eec7d8R121 ... it COULD be removed but may actually still be useful.

PR #29, opened 2019-07-24
Fix os.File.MoveToLocation()

Roosh513:
The underlying os.Rename() function was throwing errors when attempting to move files to directories that do not exist. This change will ensure the directory exists before doing the Rename() call.

### DIFF
--- a/backend/os/file.go
+++ b/backend/os/file.go
@@ -164,6 +164,9 @@ func (f *File) MoveToFile(file vfs.File) error {
 func (f *File) MoveToLocation(location vfs.Location) (vfs.File, error) {
 	// handle native os move/rename
 	if location.FileSystem().Scheme() == Scheme {
+		if err := ensureDir(location); err != nil {
+			return nil, err
+		}
 		err := os.Rename(f.Path(), path.Join(location.Path(), f.Name()))
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


backend/os/file.go
```diff
@@ -261,3 +264,14 @@ func (f *File) openFile() (*os.File, error) {
 	f.file = file
 	return file, err
 }
+
+func ensureDir(location vfs.Location) error {
+	if exists, err := location.Exists(); err != nil {
+		return err
+	} else if !exists {
+		if err := os.MkdirAll(location.Path(), os.ModeDir|0777); err != nil {
```
garrypolley: The `0777` seems too broad for creating a new directory and code path. Not sure of a great solution though that isn't a fair amount of more work. Like looking at the target to move and seeing it's permissions.

jkc: I think this is fine as is, unless we want to either make limited arbitrary permissions choices for end users, or actually make some common subset of all the various underlying file system's permissions models available in the interfaces. Though there is very little common ground between the permissions models of the various file systems, to the point that I wonder if you need permissions controls outside of the inherent limitations of the user running the code (or the remote user for a remote file system) you may be better off using the underlying implementation's library for that (golang's os package in this case), rather than vfs. 

It is a tricky issue for sure...

PR #33, opened 2019-09-09
Adding sftp backend

funkyshu:


backend/sftp/file_test.go
```diff
@@ -0,0 +1,660 @@
...
+	f, err := file.openFile(os.O_RDONLY)
+	ts.NoError(err, "no error expected")
+	b, err := ioutil.ReadAll(f)
+	ts.NoError(err, "no error expected")
+	ts.Equal(string(b), "file 1", "mock returned")
+
+	//file not open, open for read
+	file.sftpfile = nil
+	file.opener = func(c Client, p string, f int) (SFTPFile, error) { return file1, nil }
+	file1.Seek(0, 0) //reset file
```
golangcibot: Error return value of `file1.Seek` is not checked (from `errcheck`)

backend/sftp/file_test.go
```diff
@@ -0,0 +1,660 @@
...
+	file.opener = nil
+	client.On("OpenFile", filepath, os.O_RDONLY).Return(&sftp.File{}, nil)
+	f, err = file.openFile(os.O_RDONLY)
+	ts.NoError(err, "no error expected")
+	ts.IsType(&sftp.File{}, f, "type check")
+
+	//file not open, open for create/write
+	file.sftpfile = nil
+	file.opener = func(c Client, p string, f int) (SFTPFile, error) { return file1, nil }
+	file1.Seek(0, 0) //reset file
```
golangcibot: Error return value of `file1.Seek` is not checked (from `errcheck`)

backend/sftp/options_test.go
```diff
@@ -0,0 +1,595 @@
...
+		for k, v := range tmpMap {
+			o.NoError(os.Setenv(k, v))
+		}
+	}
+}
+
+type getClientTest struct {
+	options    Options
+	authority  utils.Authority
+	envVars    map[string]string
```
golangcibot: `envVars` is unused (from `structcheck`)

backend/sftp/file.go
```diff
@@ -0,0 +1,341 @@
...
+func (f *File) String() string {
+	return f.URI()
+}
+
+/*
+	Private helper functions
+*/
+
+// openFile wrapper allows us to inject a file opener (for mocking) vs the defaultOpenFile.
+// This solves for the fact that
```
moezeid: unfinished thought?

funkyshu: yep..should be fixed now.

backend/sftp/file.go
```diff
@@ -0,0 +1,341 @@
...
+
+	if err := utils.TouchCopy(file, f); err != nil {
+		return err
+	}
+	//Close target to flush and ensure that cursor isn't at the end of the file when the caller reopens for read
+	if cerr := file.Close(); cerr != nil {
+		return cerr
+	}
+	//Close file (f) reader
+	return f.Close()
```
Roosh513: Shouldn't this be up to the implementer using the API? Closing internally may do something unintended, such as writing file contents remotely before file operations are complete after the copy.

```
myFile, _ := fs.NewFile(...)
otherFile, _ := fs.NewFile(...)
myFile.CopyToFile(otherFile)
myFile.Write([]byte("moar data!"))
myFile.Close()
```
This would result in the follow-up write to overwrite existing data rather than appending, and I'd have no idea unless I dug into the API source.

backend/sftp/file.go
```diff
@@ -0,0 +1,341 @@
...
+
+	if err := utils.TouchCopy(newFile, f); err != nil {
+		return nil, err
+	}
+	//Close target file to flush and ensure that cursor isn't at the end of the file when the caller reopens for read
+	if cerr := newFile.Close(); cerr != nil {
+		return nil, cerr
+	}
+	//Close file (f) reader
+	if cerr := f.Close(); cerr != nil {
```
Roosh513: Same here, this is making decisions for the API user around the state of the source file. Maybe this is just keeping things consistent with the other implementations, but we should probably clarify the behavior in docs.

docs/sftp.md
```diff
@@ -0,0 +1,539 @@
...
+User may only be set in the [URI authority](https://tools.ietf.org/html/rfc3986#section-3.2) section (Volume in vfs parlance).
+
+     scheme             host
+     __/             ___/____  port
+    /  \            /        \ /\
+    sftp://someuser@server.com:22/path/to/file.txt
+    	   \____________________/ \______________/
+    	   \______/       \               \
+    		  /     authority section    path
+    	 username       (Volume)
```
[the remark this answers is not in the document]
jkc: It is the state of the art top secret model.

backend/sftp/file.go
funkyshu: In this case the CopyToFile, "opening" the file wasn't initiated by the buyer. We close internally to make this an atomic function to the user.  In your scenario, CopyToFile will jhave opened and closed the file and Write will open again, necessitating an explicit close.  Basically, we required a close on Read, Write, and Seek.

backend/sftp/file.go
funkyshu: Same response here.

backend/sftp/file.go
funkyshu: Grrr... i wrote more expansive reply on this but lost it.  Basically, I said I'd add more docs to clarify to for these functions, explaining that we close both source and target files.  This is generally consistent with unix command line usage of `cp` or `mv` which was the behavior model.  Also noted that you can continue to append with an additional seek-to-end call 
```go
myFile, _ := fs.NewFile(...)
otherFile, _ := fs.NewFile(...)
myFile.CopyToFile(otherFile)
myFile.Seek(0,2)
myFile.Write([]byte("moar data!"))
myFile.Close()
```
See the latest commit.

Issue #40, opened 2020-03-24
os backend write doesn't truncate existing file


funkyshu:
VFS os backend has a bug where files opened for write are not truncated first.
      given:
         THIS IS SOME TEXT ALREADY IN FILE
      open file for write and write "this is my text" would return
         this is my textXT ALREADY IN FILE
The os file should be (re)opened with O_TRUNC when writing to ensure it gets truncted OR at least call
the sytem packages *os.File's Truncate() method when called (for an initial write)
This bug is not an issue when the file being written to is smaller than the reader.

PR #41, opened 2020-04-24
EP-3173 | Ensure that writing truncates existing file. Fizes #40.

thatsmysky:
Ensure that os backend truncates existing file when writing. 

https://github.com/C2FO/vfs/issues/40

### DIFF
--- a/backend/os/file.go
+++ b/backend/os/file.go
@@ -387,7 +387,7 @@ func (f *File) copyToLocalTempReader() (*os.File, error) {
 	if _, err = openFunc(f.Path()); err != nil {
 		return nil, err
 	}
-	// todo: editing in place logic/appending logic
+	// todo: editing in place logic/appending logic (see issue #42)
 	//if _, err := io.Copy(tmpFile, f.file); err != nil {
 	//	return nil, err
 	//}
--- a/backend/os/file.go
+++ b/backend/os/file.go
@@ -1,8 +1,8 @@
 package os
 
 import (
-	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
@@ -12,11 +12,17 @@ import (
 	"github.com/c2fo/vfs/v5/utils"
 )
 
+type opener func(filePath string) (*os.File, error)
+
 //File implements vfs.File interface for os fs.
 type File struct {
-	file     *os.File
-	name     string
-	location vfs.Location
+	file        *os.File
+	name        string
+	location    vfs.Location
+	cursorPos   int64
+	tempFile    *os.File
+	useTempFile bool
+	fileOpener  opener
 }
 
 // Delete unlinks the file returning any error or nil.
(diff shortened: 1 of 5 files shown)


backend/os/file.go
```diff
@@ -275,3 +323,53 @@ func ensureDir(location vfs.Location) error {
...
+func (f *File) copyToLocalTempReader() (*os.File, error) {
+	tmpFile, err := ioutil.TempFile("", fmt.Sprintf("%s.%d", f.Name(), time.Now().UnixNano()))
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err = f.openFile(); err != nil {
+		return nil, err
+	}
+	// todo: editing in place logic/appending logic
```
jkc, line 364: Should these comments still be here, or is there a reason the todo should be handled after merge?

backend/os/file.go
```diff
@@ -61,6 +64,23 @@ func (f *File) Size() (uint64, error) {
...
+	// os.Rename() (replace) temp file to file
+	f.useTempFile = false
+	f.cursorPos = 0
+	if f.tempFile != nil {
+		err := f.tempFile.Close()
+		if err != nil {
+			return err
+		}
+
+		err = os.Rename(f.tempFile.Name(), f.file.Name())
```
jkc, line 78: 👍 Better solution than copying the temp file over a RDRW|TRUNC file descriptor.

backend/os/file.go
thatsmysky: It is part of a larger change that will likely be the next fix on this file, but too out of scope for the current fix.  Left the code there and commented instead of deleting it and rewriting it later.

Roosh513: Might be good to log an issue related to this TODO

backend/os/fileSystem.go
```diff
@@ -13,7 +14,10 @@ const Scheme = "file"
 const name = "os"
 
 // FileSystem implements vfs.Filesystem for the OS file system.
-type FileSystem struct{}
+type FileSystem struct {
+	client  s3iface.S3API
```
Roosh513, line 18: This was probably a copy/paste mistake. I'm guessing the OS implementation doesn't need the S3 interface.

backend/os/file.go
thatsmysky: im not sure how to do that, but I agree it would be a good thing to do

backend/os/fileSystem.go
thatsmysky: Absolutely overzealous copy and paste. Before I go and remove the change entirely (tests still pass) is there any reason to keep the vfs.Options portion?

backend/os/file.go
funkyshu: Now that you added issue #42, can you just put that your today comment?  something like this should suffice:
```
// todo: editing in place logic/appending logic (see issue #42)
```

PR #43, opened 2020-05-11
Fixed issue related to https://github.com/pkg/sftp/pull/310/files

Roosh513:
While attempting to connect to an SFTP server hosted on Amazon AWS, the following error is thrown:

> Cannot open file in mode: WRITE|CREATE|READ" (SSH_FX_OP_UNSUPPORTED)

This is related to https://github.com/pkg/sftp/issues/305

### DIFF
--- a/backend/sftp/file.go
+++ b/backend/sftp/file.go
@@ -76,7 +76,7 @@ func (f *File) Touch() error {
 	}
 
 	if !exists {
-		file, err := f.openFile(os.O_RDWR | os.O_CREATE)
+		file, err := f.openFile(os.O_WRONLY | os.O_CREATE)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


backend/sftp/file.go
```diff
@@ -252,7 +252,7 @@ func (f *File) Seek(offset int64, whence int) (int64, error) {
 // Write calls the underlying sftp.File Write.
 func (f *File) Write(data []byte) (res int, err error) {
 
-	sftpfile, err := f.openFile(os.O_RDWR | os.O_CREATE)
+	sftpfile, err := f.openFile(os.O_WRONLY | os.O_CREATE)
```
funkyshu, line 255: I think this can work but this won't address the same issue if we try to seek. Seek is the tricky one because it could be for both read or write.  The way we were approaching that with the append vs overwrite issue is doing a remote-to-local copy before the first write, then acting on the temp, until close write back to remote.  Here we're doing direct remote writes/seek/etc.

PR #46, opened 2020-07-06
bug fixes

### LINKED ISSUE #21: VFS: OS fs uses a reference to a file's location when calling "ChangeDir"
File structs in the OS fs have a member location.  File's **"Location()"** function returns the member location by simply saying _**"return file.location"**_

This becomes a problem with ChangeDir.  Say we have a file named **"foo.txt"** that has a location of _**"/home/test_files/subDir/**"_ and we want to call _**ChangeDir("relPath")**_ on that location. We would do so by saying _**"file.Location().ChangeDir("relPath")**"_ and the new location would be 
**"/home/test_files/subDir/relPath/"**.  This is an issue because the call to "Location()" returns that file's member location.  
The consequence would be that although we never moved the file, its path will have changed.  Calling _**file.Path()**_ would yield _**"home/test_files/subdir/relPath/foo.txt"**,_ which is incorrect.  
The filesystem DOES know that that location does not exist, nevertheless that is where it says the file is located.  A potential solution is returning a copy of the current location if _**"file.Location()"**_ is called

### LINKED ISSUE #44: S3 from Exists() check thrown away here.
https://github.com/C2FO/vfs/blob/8b1d1e1000bc162af5125d500d3d0e6416e78b5e/backend/s3/file.go#L458

The way the error handling is written here, if Exists() results in an error, the specific error from S3 is thrown away, and replaced with "unable to check for file on S3". While this message is fine, we should include the error returned from the API along with this, to give the end user context to troubleshoot what is going wrong. 

As an example, when the (s3) user running the code doesn't have GetObject permissions, this will return a Forbidden error. The code had to be modified locally to discover that.

### LINKED ISSUE #45: SFTP ASCII diagram doesn't align in markdown
Looks like there are tabs vs space
in 
https://github.com/C2FO/vfs/blob/master/docs/sftp.md
and 
https://github.com/C2FO/vfs/blob/master/backend/sftp/doc.go#L95-L102

funkyshu:
### Fixed
- fixes #21 - OS file.Location().ChangeDir(something/) should not affect location
- fixes #44 - make S3 waitUntilFileExists Exists error more descriptive of root error
- fixes #45 - fix ASCII diagram alignment for SFTP doc
### Added
- Improved test coverage for os.FileSystem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.5.4] - 2020-07-06
+### Fixed
+- fixes #44 - make S3 waitUntilFileExists Exists error more descriptive of root error
+- fixes #45 - ASCII diagram alignment for SFTP doc
+
 ## [5.5.3] - 2020-05-11
 ### Fixed
 - Addresses an issue where writes/touch calls on the vfs.File backend did not work properly on AWS-hosted SFTP environments. (See https://github.com/pkg/sftp/pull/310/files)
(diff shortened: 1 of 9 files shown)


funkyshu:
I probably should've made these separate commits but once i did one, I just wanted to knock out the outstanding issues.

PR #54, opened 2021-02-01
Cross dev link os move bug

funkyshu:
fixes#48 bug where 'invalid cross-device link' error occurs when renaming os files that are found on different devices/volumes including temp dirs


Roosh513 (approved):

thatsmysky:
is it possible to add tests so we don't reintroduce this bug in future changes?

thatsmysky (approved):

PR #55, opened 2021-02-05
Fixes #53 bug where s3 file tests were failing due to a defered remov…

### LINKED ISSUE #53: S3 file Close defers the close of the temp file but deletes the file causing an error on Windows
The function Close defers the close of the temp file but does an os.Remove on the file.
This causes the following error to occur
"internal/syscall/windows.ERROR_SHARING_VIOLATION (32)"

This is the defer line that should not wait to close the file otherwise the sharing violation occurs on windows.
https://github.com/C2FO/vfs/blob/5a37f1aa8d9111bfae552de195d30f747bf8ae3f/backend/s3/file.go#L185

Instead of deferring the file closed it should close and then removed.

Error occurs after the os.Remove() because the file has not been closed.
This link is the line where the error can be seen.
https://github.com/C2FO/vfs/blob/5a37f1aa8d9111bfae552de195d30f747bf8ae3f/backend/s3/file.go#L188

Actual code looks like
------------------------
// Close cleans up underlying mechanisms for reading from and writing to the file. Closes and removes the
// local temp file, and triggers a write to s3 of anything in the f.writeBuffer if it has been created.
func (f *File) Close() error {

	if f.tempFile != nil {
		defer func() { _ = f.tempFile.Close() }()

		err := os.Remove(f.tempFile.Name())
		if err != nil && !os.IsNotExist(err) {
			return err
		}

		f.tempFile = nil
	}

Should change the code to look something like:
-----------------------------------------------------
// Close cleans up underlying mechanisms for reading from and writing to the file. Closes and removes the
// local temp file, and triggers a write to s3 of anything in the f.writeBuffer if it has been created.
func (f *File) Close() error {

	if f.tempFile != nil {
		_ = f.tempFile.Close()    <-- This line should no be defer'd. 

		err := os.Remove(f.tempFile.Name())
		if err != nil && !os.IsNotExist(err) {
			return err
		}

		f.tempFile = nil
	}

Thank you for your time.

funkyshu:
…al of a temp file.

### DIFF
--- a/backend/s3/file.go
+++ b/backend/s3/file.go
@@ -182,9 +182,12 @@ func (f *File) Delete() error {
 func (f *File) Close() error {
 
 	if f.tempFile != nil {
-		defer func() { _ = f.tempFile.Close() }()
+		err := f.tempFile.Close()
+		if err != nil {
+			return err
+		}
 
-		err := os.Remove(f.tempFile.Name())
+		err = os.Remove(f.tempFile.Name())
 		if err != nil && !os.IsNotExist(err) {
 			return err
 		}


backend/s3/file.go
```diff
@@ -182,9 +182,12 @@ func (f *File) Delete() error {
...
+		err := f.tempFile.Close()
+		if err != nil {
+			return err
+		}
 
-		err := os.Remove(f.tempFile.Name())
+		err = os.Remove(f.tempFile.Name())
 		if err != nil && !os.IsNotExist(err) {
 			return err
 		}
```
funkyshu, line 193: Note, this was a windows-only issue.  There are test failures in Windows in other backends.  Will add issues for them as well updating CI to include Windows testing.

Roosh513 (approved):

PR #57, opened 2021-04-16
EP-5510 -> master | S3 VFS backend throws an error when you open then close a file.

shivanigaurh-c2fo:
Fix waitUntiFileExists bug when file is created but now write occurs.

### DIFF
--- a/backend/s3/file.go
+++ b/backend/s3/file.go
@@ -211,13 +211,9 @@ func (f *File) Close() error {
 		}
 
 		f.writeBuffer = nil
-		err = waitUntilFileExists(f, 5)
-		if err != nil {
-			return err
-		}
-	}
+		return waitUntilFileExists(f, 5)
 
-	f.writeBuffer = nil
+	}
 	return nil
 }
 
--- a/backend/s3/file.go
+++ b/backend/s3/file.go
@@ -209,12 +209,14 @@ func (f *File) Close() error {
 		if err != nil {
 			return err
 		}
+
+		f.writeBuffer = nil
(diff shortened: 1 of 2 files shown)


backend/s3/file.go
```diff
@@ -209,11 +209,12 @@ func (f *File) Close() error {
 		if err != nil {
 			return err
 		}
+		return waitUntilFileExists(f, 5)
```
funkyshu, line 212: we should check for the error and if return it, if any. But on success, we still need to call the line below where we set `f.writeBuffer = nil` because the buffer should be emptied any time we close() a file.

funkyshu (changes requested):

backend/s3/file.go
```diff
@@ -209,11 +209,16 @@ func (f *File) Close() error {
...
+		if err != nil {
+			return err
+		}
 	}
 
 	f.writeBuffer = nil
-
-	return waitUntilFileExists(f, 5)
+	return nil
 }
```
funkyshu, line 222: setting writeBuffer to nil only needs to happen inside the `if f.writeBuffer != nil {` block (because otherwise it would already be nil).  I think you may have had it like this before but I think we can simplify this to:
```go
		}

		f.writeBuffer = nil
		return waitUntilFileExists(f, 5)
	}

	return nil
}
```

funkyshu (changes requested):

atulbhadouria (approved):

funkyshu (approved):

PR #68, opened 2021-05-05
Add OAuth token based authentication

dmcilvain44:
Add the ability to authenticate with service accounts via OAuth tokens.

### DIFF
--- a/backend/azure/client.go
+++ b/backend/azure/client.go
@@ -157,7 +157,7 @@ func (a *DefaultClient) Copy(srcFile, tgtFile vfs.File) error {
 		return nil
 	}
 
-	return fmt.Errorf("Copy failed.  ERROR[%s]", resp.ErrorCode())
+	return fmt.Errorf("copy failed ERROR[%s]", resp.ErrorCode())
 }
 
 // List will return a listing of the contents of the given location.  Each item in the list will contain the full key
@@ -180,8 +180,8 @@ func (a *DefaultClient) List(l vfs.Location) ([]string, error) {
 
 		marker = listBlob.NextMarker
 
-		for _, blobInfo := range listBlob.Segment.BlobItems {
-			list = append(list, blobInfo.Name)
+		for i := range listBlob.Segment.BlobItems {
+			list = append(list, listBlob.Segment.BlobItems[i].Name)
 		}
 	}
 	return list, nil
--- a/backend/azure/client.go
+++ b/backend/azure/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"log"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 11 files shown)


backend/azure/client.go
```diff
@@ -183,7 +175,7 @@ func (a *ClientImpl) List(l vfs.Location) ([]string, error) {
 		listBlob, err := containerURL.ListBlobsHierarchySegment(ctx, marker, "/",
 			azblob.ListBlobsSegmentOptions{Prefix: utils.RemoveLeadingSlash(l.Path())})
 		if err != nil {
-			log.Fatal("Unable to list blobs.  error: " + err.Error())
+			return []string{}, err
```
thatsmysky, line 178: no more error log or does it get caught in the return?

backend/azure/options_test.go
```diff
@@ -0,0 +1,81 @@
...
+	o := NewOptions()
+	s.NotNil(o, "when NewOptions returns an error we expect to get a nill options struct")
+}
+
+func (s *OptionsTestSuite) TestCredentials_ServiceAccount() {
+	options := Options{
+		AccountName:            "foo",
+		TenantID:               "foo",
+		ClientID:               "foo",
+		ClientSecret:           "foo",
```
thatsmysky, line 30: Do we know what these values would be expected to be? To test that the numbers of the ID etc will parse correctly?

thatsmysky (changes requested):

backend/azure/client.go
dmcilvain44: This was a defect in the original PR.  The log.Fatal would cause a panic if this error condition occurred.

backend/azure/client.go
```diff
@@ -19,53 +18,46 @@ import (
 // The Client interface contains methods that perform specific operations to Azure Blob Storage.  This interface is
 // here so we can write mocks over the actual functionality.
 type Client interface {
-	// Properties should return a BlobProperties struct for the blob specified by locationURI, and filePath.  If the
-	// blob is not found an error should be returned.
 	Properties(locationURI, filePath string) (*BlobProperties, error)
```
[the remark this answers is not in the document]
dmcilvain44: This was the result of a bad merge.  I had these changes stashed before I added the comments and they got removed when I merged the stash back.

backend/azure/options_test.go
dmcilvain44: All we know is that the values need to be valid credentials.  The credentials are validated on authentication.

backend/azure/options.go
```diff
@@ -10,6 +16,47 @@ type Options struct {
...
+	// AzureEnvName holds the name for the Azure environment
+	AzureEnvName string
+
 	// RetryFunc holds the retry function
 	RetryFunc vfs.Retry
+
+	tokenCredentialFactory TokenCredentialFactory
+}
+
+func NewOptions() *Options {
```
funkyshu, line 37: Weird that golangci-lint isn't reporting the golint returns but we should add comments to any exported functions/types/vars etc.

funkyshu (approved):

thatsmysky (approved):

PR #72, opened 2021-06-20
Fix gs.Location.ListByPrefix and Exists

### LINKED ISSUE #71: Listing a Google Cloud Storage bucket does not return things in the root of the bucket.
If I use `Location.List()` ([this](https://pkg.go.dev/github.com/c2fo/vfs/v5@v5.5.7/backend/gs#Location.List)) on a google cloud storage location which is just a bucket with files directly in the bucket and not in any "folder" it does not find anything.

pseudocode:

```go
gsLocation, _ := vfssimple.NewLocation("gs://bucket/")
list, _ := gsLocation.List()
# list will be empty if all objects in the bucket is directly in the root without any folders.
```

The problem is, as far as I can tell, that the query prefix in this case is `/`, while files directly in a bucket has no prefix.

Making this change fixes it:

```go
$ git diff vendor/
diff --git a/suggestions-go/vendor/github.com/c2fo/vfs/v5/backend/gs/location.go b/suggestions-go/vendor/github.com/c2fo/vfs/v5/backend/gs/location.go
index 7a0abca..5354ab0 100644
--- a/suggestions-go/vendor/github.com/c2fo/vfs/v5/backend/gs/location.go
+++ b/suggestions-go/vendor/github.com/c2fo/vfs/v5/backend/gs/location.go
@@ -38,6 +38,9 @@ func (l *Location) ListByPrefix(filenamePrefix string) ([]string, error) {
        if filenamePrefix == "" {
                prefix = utils.EnsureTrailingSlash(prefix)
        }
+       if prefix == "/" {
+               prefix = ""
+       }
        d := path.Dir(prefix)
        q := &storage.Query{
                Delimiter: "/",
```

aucampia:
gs.Location.ListByPrefix only worked for blank prefixes
that are not in the root of the bucket.

gs.Location.Exists checked against object does not exists
instead of bucket does not exist.

Also added a bunch of unit tests for gs.Location, mostly copied
from s3.

The unit tests make use of
github.com/fsouza/fake-gcs-server/fakestorage.

Fixes #71


backend/gs/location.go
```diff
@@ -34,10 +34,7 @@ func (l *Location) List() ([]string, error) {
 // ListByPrefix returns a slice of file base names and any error, if any
 // List functions return only file basenames
 func (l *Location) ListByPrefix(filenamePrefix string) ([]string, error) {
-	prefix := utils.RemoveLeadingSlash(path.Join(l.prefix, filenamePrefix))
-	if filenamePrefix == "" {
```
aucampia, line 38: Query strings should always end with a slash if they are to work.

funkyshu:
Merged.  Thanks for that patch and tests.  I'll get this tagged in the next day or two.

PR #75, opened 2021-06-21
Removed need to call Size() in TouchCopy. Also reduced signature to o…

### LINKED ISSUE #74: utils.TouchCopy() unnecessarily calls file.Size()
To ensure uniform behavior, VFS expects that copying or moving an empty file from one place to another results in an empty file at the target location.  However, in some implementations of VFS (like S3/GCS), copying a empty file using io.Copy will not result in any action at the target.  To this end we created a utility function TouchCopy() that works exactly like io.Copy except that it does special handling for copying empty files to ensure they result in an new empty target file.

The current implementation checks the size of the file first.  If the size = 0 bytes, special handling ensures that new empty file gets created (by calling Write() on an empty byte array).  Otherwise, io.Copy is called. This is inefficient since io.Copy already returns the amount of bytes copied.  If we try io.Copy first and check its bytes written count value, we no long need to call Size() on the vfs source.  Size on `os` does a stat while cloud systems do a HEAD operation on the object.

We should make TouchCopy first try io.Copy().  Now if we check bytes written, we can return on non-zero or call src.Write([]byte{}).  This removes the need for calling Size().  Indeed it removes the need for TouchCopy to required vfs.Files at all.  We only need io.Reader and io.Writer, just like io.Copy().

funkyshu:
…nly need io.Reader and io.Writer, rather than vfs.File.

fixes #74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
+### Fixed
+- fixed #71 Listing a Google Cloud Storage bucket does not return things in the root of the bucket
+### Performance
+- fixed #74 utils.TouchCopy() unnecessarily calls file.Size()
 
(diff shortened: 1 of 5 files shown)


Roosh513 (approved):
Looks like tests are failing

PR #77, opened 2021-06-23
Add some unit tests for gs/file.go

aucampia:
I am mainly doing this because I was experiencing some bug with
MoveToFile in v5.5.7 which is no longer present in the master branch.

### DIFF
--- a/backend/gs/file_test.go
+++ b/backend/gs/file_test.go
@@ -1,0 +1,211 @@
+package gs
+
+import (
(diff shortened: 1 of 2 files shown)


Roosh513 (approved):
Thank you for adding these! I like the fakestorage library you chose. It makes for some clean tests on the `gs` backend.

funkyshu (approved):

PR #79, opened 2021-06-24
gs backend: Use a clean file handle when copying files using io.Reader

### LINKED ISSUE #78: Using gs.File.CopyToFile after reading the file results in 0 byte file at the destination
This is because the same reader is used for copying and for reading, so by the time you get to CopyToFile the reader is already at the end of the file.

Probably the simplest solution to this fix will be to just Seek(0,0) before doing the copy.

As a workaround the seek can be done in user code.

Any inputs on possible resolutions is welcome, I think it would be good to just call Seek(0, 0) and I will prepare a patch for that.

aucampia:
Under some circumstances gs.File.CopyToFile copies using f.tempFile as a
io.Reader. The problem with this is that if the cursor is not at
position 0, copying using io.Reader will not copy the whole file.

This change solves this problem by using a clean reader when copying
using io.Reader so that the copy is not affected by the reader cursor
position.

Fixes #78

### DIFF
--- a/backend/gs/file.go
+++ b/backend/gs/file.go
@@ -154,7 +154,13 @@ func (f *File) CopyToFile(file vfs.File) error {
 		}
 	}
 
-	if err := utils.TouchCopy(file, f); err != nil {
+	// Get a clean handle so that the cursor position does not affect what gets copied.
+	handle, err := f.newTempFileHandle()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


funkyshu:
Closing in favor of returning an error on any Move/Copy function if cursor was not at 0,0.

PR #81, opened 2021-07-20
Add support of key exchange algorithm as an option

atulbhadouria:
fixes #81


vikram-c2fo (approved):

funkyshu (changes requested):
This looks good.  Can you update https://github.com/C2FO/vfs/blob/master/docs/sftp.md#type-options to reflect the change in the Options struct?
Can you also update the Unreleased sectionof https://github.com/C2FO/vfs/blob/master/CHANGELOG.md .  This change would fall under an `Added` section.

Roosh513 (approved):

funkyshu (approved):

PR #83, opened 2021-07-26
fixed #82 Return error when CopyTo/MoveTo functions are called when S…

### LINKED ISSUE #82: Ensure copyto/moveto seek position validation in all backends
GCS backend added a error in #78 when seek cursor is not at 0 when calling MoveTo/CopyTo functions. This ensure that no partial files get copied/moved.  We should apply this to all backends.

funkyshu:
fixes #82 
Return error when CopyTo/MoveTo functions are called when Seek offset is not (0,0) for all backends, not just GCS

### DIFF
--- a/backend/azure/file.go
+++ b/backend/azure/file.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/azure-storage-blob-go/azblob"
 
 	"github.com/c2fo/vfs/v5"
+	"github.com/c2fo/vfs/v5/backend"
 	"github.com/c2fo/vfs/v5/utils"
 )
 
@@ -29,7 +30,7 @@ type File struct {
 func (f *File) Close() error {
 	if f.tempFile != nil {
 		defer func() {
-			f.tempFile.Close()
+			_ = f.tempFile.Close()
 			f.tempFile = nil
 			f.isDirty = false
 		}()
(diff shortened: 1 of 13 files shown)


backend/azure/file.go
```diff
@@ -29,7 +30,7 @@ type File struct {
 func (f *File) Close() error {
 	if f.tempFile != nil {
 		defer func() {
-			f.tempFile.Close()
+			_ = f.tempFile.Close()
```
thatsmysky, line 33: we would never want to do anything with these errors?

thatsmysky (approved):

Roosh513 (approved):

backend/azure/file.go
funkyshu: good question... here i'm not changing any behavior except making it explicit what we're doing.  Close() is really the hardest thing to do right.  Either you run Close() in a defer or you rerun it at every exit point of the function.  But each exit point of the function either returns an error already or nil (no error).  Unfortunately, we can't return a value (or error) from inside defer.  Furthermore, we can't really return 2 errors without concatenating them somehow.  
The particular quandary of handling errors on Close() doesn't really seem to have a community standard or idiomatic approach.  We made a stab at it early using a error type that allowed us to append errors to AND utilized named error returns so that both the normal code and Close() defer closure would update the error.  You can see the code at https://github.com/C2FO/vfs/blob/8558f12c8f4cb571cd32a610936f0c1d48aa737b/errors.go and example usage at https://github.com/C2FO/vfs/blob/8558f12c8f4cb571cd32a610936f0c1d48aa737b/errors_example_test.go.  However we soon found this a cumbersome and kludgy solution with very little gain.  Close() is the one return value that we regularly ignore.
I'm all for a better solution though!

dmcilvain44 (approved):

juanri0s (approved):

katiehrenchir (approved):

PR #85, opened 2021-08-21
Add sftp client close 

### LINKED ISSUE #84: sftp backend never closes open sftp connections
In a log lived application, when a vfs sftp backend filesystem falls out of scope, it doesn't disconnect the underlying ssh connection to remote server.  The connections are only severed by restarting the either the calling application or the sftp server.   

As a user of the vfs library, I expect that once I've completed my SFTP operations, I have a way to disconnect my ssh connection to the server.

We should ensure that:
1. A user has a way to explicitly disconnect the connection from a sftp backed to a server
2. sftp backend auto-disconnects after 10 seconds of disuse.  This value should be configurable via options.

funkyshu:
fixed #84 where sftp connections were never disconnected in a long-lived app.  
Added Close to sftp backend Client interface.  
Close client automatically 10 seconds (configurable in Options) after connecting unless reset by calling some server request action.


backend/sftp/fileSystem.go
```diff
@@ -89,24 +95,55 @@ func (fs *FileSystem) Scheme() string {
...
+		}
+	}
+
+	fs.connTimer = time.AfterFunc(time.Duration(aliveSec)*time.Second, func() {
+		// close connection and nil-ify client to force lazy reconnect
+		if fs.sftpclient != nil {
+			_ = fs.sftpclient.Close()
+			fs.sftpclient = nil
+		}
+	})
```
funkyshu, line 135: This ends up being the meat of the changes.  This timer will disconnect after 10 seconds (default) unless a new request to the server is made (read, write, list, etc), in which case the timer will reset.  
If a timer expires and closes a connection, any subsequent server request action will request a new client, reconnecting and starting new timer.
In this way, a user need not explicit call Close on a client since it will simply timeout anyway.

arnab-chaudhuri-c2fo (approved):
Looks like reasonable changes based on my understanding how VFS works.

atulbhadouria (approved):

PR #95, opened 2021-09-10
Ep 7935 - S3 Multipart Download

arnab-chaudhuri-c2fo:
Implemented S3 Downloader to multipart download files on initial read calls. Updated test suite to mock S3 Downloader

```
Going to go with 32MB Chunks with 5 Processes Default. Results Below:
1 - 226/366 = 38% Faster Than S3 Get Objects (Single Socket Download)
366/226 = S3 Get Object 162% slower than S3 Downloader (Parallel Multipart Download)

Master Results S3 Get Object:
BenchmarkRead_AWS_File_100MB-16               10        4595231254 ns/op
BenchmarkRead_AWS_File_1GB-16                 10        36590394191 ns/op

Read with S3 Download Manager 5 MB chunks 5 Processes:
BenchmarkRead_AWS_File_100MB-16               10        4113697995 ns/op
BenchmarkRead_AWS_File_1GB-16                 10        22814272873 ns/op

Read with S# Download Manager 32 MB chunks 5 Processes:
BenchmarkRead_AWS_File_100MB-16               10        4151013363 ns/op
BenchmarkRead_AWS_File_1GB-16                 10        22638071967 ns/op
```


backend/s3/file.go
```diff
@@ -29,6 +31,14 @@ type File struct {
 	writeBuffer *bytes.Buffer
 }
 
+type S3Downloader interface {
```
seeroush, line 34: For completeness, this should include godoc comments.

seeroush (changes requested):
In general, this change looks good. Have you attempted to run the integration test? It's the only current manual test step outside of the CI job since it depends on environment variables to run.

funkyshu (approved):
Nice work!

funkyshu (changes requested):
One last thing, please update the CHANGELOG.md file.
We'll add an automated check for this soon.

backend/s3/file.go
arnab-chaudhuri-c2fo: @seeroush 
Something like this work?

**// Interface needed to mock S3 Downloader data access object in integration tests**

arnab-chaudhuri-c2fo: As a side note as part of the whole refactor file options things. I do think it could be beneficial to use dependency injection for the s3 downloader to avoid having to do some of the abstractions needed to mock things out with the current constraints.

juanri0s (approved):

backend/s3/file.go
arnab-chaudhuri-c2fo: Looks like John ran the integration tests and things passed.

funkyshu (approved):

PR #104, opened 2021-11-03
Fix autodisconnect

### LINKED ISSUE #103: sftp backend only auto-disconnects sftp subsystem, not ssh connection
`sftp` backend automatically disconnects when no command has been issued in the last 10 seconds (default).  However, it appears to only close the SFTP subsystem session, not the underlying SSH connection.  This is not a problem when the application closes but is definitely a problem on long-running apps.  Memory may eventually run out or ssh service max connections may be reached.

If you run the following test app, the server side (at least on OpenSSH server), every connection will show the main connection `sshd: <username> [priv]`, and two forked processes `sshd: <username>@notty` and `sshd: <username>@internal-sftp`.  10 second after each connection was mad the subsytem `@internal-sftp` process goes away but leaves the other two.
```go
package main

import (
	"fmt"

	"github.com/c2fo/vfs/v6/backend/sftp"
)

func main() {
        // 10 separate connections are made
	for i := 1; i < 11; i++ {
		err := list()
		if err != nil {
			panic(err)
		}
	}
	fmt.Println("DONE")

        // blocking -- ctrl-c when done
	select {}
}

func list() error {

	fs := &sftp.FileSystem{}
	fs = fs.WithOptions(sftp.Options{Password: "mySecretPassword"})
	loc, err := fs.NewLocation("myusername@sftp.myserver.com", "/home/myusername/")
	if err != nil {
		return err
	}
	fmt.Printf("checking %s\n", loc.URI())

        // connection is made here
	files, err := loc.List()
	if err != nil {
		return err
	}
	fmt.Printf("%v", files)
	return nil
}
```

Need to close the underlying ssh conn.

funkyshu:
fixes #103 

Ensures that sftp backend closed underlying SSH connection on autodisconnect, not just SFTP subsystem:
* Stores underlying ssh conn on the struct as an io.Closer.  
* Call Close when auto-disconnect timeout is triggered.

### DIFF
--- a/backend/sftp/doc.go
+++ b/backend/sftp/doc.go
@@ -133,5 +133,48 @@ Passing in multiple key exchange algorithms is supported - these are specified a
 Example:
 `"keyExchanges":["diffie-hellman-group-a256", "ecdh-sha2-nistp256"]`
 
+AutoDisconnect
+
+When dialing an TCP connection, go doesn't disconnect for you, even when the connection falls out of scope (or even when
+garbage collection is forced).  It must be explicitly closed.  Unfortunately, VFS.FileSystem has no explicit close mechanism.
+Instead, the sftp backend will automatically disconnect 10 seconds (default) after connection.  This disconnect timer is
+cancelled anytime a server-side request (like list, read, etc) is made.  Once the action is complete, a new timer will begin.
+If the timer is not interrupted by any request, it will disconnect from the server.  Any subsequent server request will
+first reconnect, the do the action.  This timer can be overridden with any number of second (zero is an immediate disconnect).
+
+Options.AutoDisconnect accepts an integer representing the number seconds before disconnecting after being idle.
+Default value is 10 seconds.
+
+Any server request action using the same underlying FileSystem (and therefore sftp client), will reset the timer.  This
+should be the most desirable behavior.
+
+	func doSFTPStuff() {
+		fs := sftp.NewFilesystem()
+		loc, err := fs.NewLocation("myuser@server.com:22", "/some/path/")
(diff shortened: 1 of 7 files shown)


thatsmysky (approved):

backend/sftp/doc.go
```diff
@@ -133,5 +133,48 @@ Passing in multiple key exchange algorithms is supported - these are specified a
 Example:
 `"keyExchanges":["diffie-hellman-group-a256", "ecdh-sha2-nistp256"]`
 
+AutoDisconnect
+
+When dialing an TCP connection, go doesn't disconnect for you, even when the connection falls out of scope (or even when
```
skduits, line 138: nit: an => a 
Also in docs/sftp.md

backend/sftp/doc.go
```diff
@@ -133,5 +133,48 @@ Passing in multiple key exchange algorithms is supported - these are specified a
...
+garbage collection is forced).  It must be explicitly closed.  Unfortunately, VFS.FileSystem has no explicit close mechanism.
+Instead, the sftp backend will automatically disconnect 10 seconds (default) after connection.  This disconnect timer is
+canceled anytime a server-side request (like list, read, etc) is made.  Once the action is complete, a new timer will begin.
+If the timer is not interrupted by any request, it will disconnect from the server.  Any subsequent server request will
+first reconnect, the do the action.  This timer can be overridden with any number of second (zero is an immediate disconnect).
```
skduits, line 143: nit: the => then; second => seconds also?
Also in docs/sftp.md

dmcilvain44 (approved):

juanri0s (approved):

PR #106, opened 2021-11-28
Fix vfssimple factory

### LINKED ISSUE #97: Resolving registered FS
I'm use vfssimple with regestered fs (s3) as shown in example: https://github.com/C2FO/vfs/blob/ab2673ff5f8388b94a8db22464de5c2f7e3e8105/docs/vfssimple.md#usage

But this example don't work without small additional:

backend.Unregister(s3.Scheme) 

Backend resolve fs by sheme\sheme+bucket\scheme+bucket+name and return first found - default registered fs. 

Maybe there is an option to disable automatic registration? If not, then it may be add to example?

funkyshu:
fixes #97 
- fixed azure helper func for vfssimple, ensuring it works on File URIs in addition to Location URIs
- fixed by updating vfssimple logic to ensure the most specific registered backend that matches a url is used, not just the first one it comes across.  Updated vfssimple docs.
- Added vfssimple tests.  Zero to 100% coverage.

### DIFF
--- a/backend/azure/fileSystem.go
+++ b/backend/azure/fileSystem.go
@@ -127,14 +127,22 @@ func init() {
 	backend.Register(Scheme, NewFileSystem())
 }
 
-// ParsePath is a utility function used by vfssiple to separate the host from the path.  The first parameter returned
+// ParsePath is a utility function used by vfssimple to separate the host from the path.  The first parameter returned
 // is the host and the second parameter is the path.
 func ParsePath(p string) (host, pth string, err error) {
 	if p == "/" {
 		return "", "", errors.New("no container specified for Azure path")
 	}
+	var isLocation bool
+	if p[len(p)-1:] == string(os.PathSeparator) {
+		isLocation = true
+	}
 	l := strings.Split(p, string(os.PathSeparator))
-	return l[1], utils.EnsureTrailingSlash(utils.EnsureLeadingSlash(path.Join(l[2:]...))), nil
+	p = utils.EnsureLeadingSlash(path.Join(l[2:]...))
+	if isLocation {
+		p = utils.EnsureTrailingSlash(p)
+	}
+	return l[1], p, nil
 }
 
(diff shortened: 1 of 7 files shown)


backend/azure/fileSystem.go
```diff
@@ -127,14 +127,22 @@ func init() {
...
+		isLocation = true
+	}
 	l := strings.Split(p, string(os.PathSeparator))
-	return l[1], utils.EnsureTrailingSlash(utils.EnsureLeadingSlash(path.Join(l[2:]...))), nil
+	p = utils.EnsureLeadingSlash(path.Join(l[2:]...))
+	if isLocation {
+		p = utils.EnsureTrailingSlash(p)
+	}
+	return l[1], p, nil
 }
```
funkyshu, line 146: No real logic change here.  Just made `EnsureTrailingSlash` conditional since this func will be used for file URIs too.

dmcilvain44 (approved):

vfssimple/vfssimple.go
```diff
@@ -31,78 +37,81 @@ func NewLocation(uri string) (vfs.Location, error) {
...
+		authority, path, err = azure.ParsePath(path)
 	}
 
-	if fs == nil {
-		err = fmt.Errorf("%s is an unsupported uri scheme", u.Scheme)
+	if u.User.String() != "" {
+		authority = fmt.Sprintf("%s@%s", u.User, u.Host)
+	}
+	// network-based schemes require authority, but not file:// or mem://
+	if authority == "" && !(scheme == "file" || scheme == "mem") {
```
seeroush, line 80: These should probably reference the constants in each of the supplied backends:

`backend.os.Scheme` and `backends.mem.Scheme`

seeroush (changes requested):
just a minor nit, but otherwise 👍

seeroush (approved):

PR #119, opened 2022-06-17
Add support for deleting file versions to s3 and gs

dhondgepooja:
### Changes Included:

- file.delete() and location.deleteFile() now support optional argument to specify delete options
- S3's implementation of file.delete() now removes all the versions of the file object if deleteOptions.DeleteAllVersions is passed in the arguments.
- GS's implementation of file.delete() now removes all the generations of the file object if deleteOptions.DeleteAllVersions is passed in the arguments.
- Other implementations of File interface, are noop for deleteOptions.DeleteAllVersions. (Azure implementation is pending)

### Design:

- In order to keep the delete interface non-passive, decided to change it to a variadic function that takes an interface as argument.
- If we wish to add additional options for delete, the new options will simply implement the options.DeleteOption interface and add any additional methods that are needed for that option.
- This approach also allows us to add support for options for other methods as well. For example, if we want to pass file permissions while creating a file, we will just create an interface like FilePermissionsOption and create implementations according to the need.
- Here is an example that shows the design https://github.com/C2FO/vfs/compare/judd-delete-options


backend/gs/file_test.go
```diff
@@ -113,6 +114,118 @@ func (ts *fileTestSuite) TestRead() {
...
+	file, err := fs.NewFile(bucketName, "/invalidObject")
+	if err != nil {
+		ts.Fail("Shouldn't fail creating new file")
+	}
+
+	err = file.Delete()
+	ts.NotNil(err, "Should return an error if gs client had error")
+}
+
+func (ts *fileTestSuite) TestDeleteRemoveAllVersions() {
```
dhondgepooja, line 182: For this test, I couldn't use more than one versions because in the fakestorage when you add second object with same name, the first object is moved under archived objects list https://github.com/fsouza/fake-gcs-server/blob/main/internal/backend/memory.go#L43 and when you do deletes on all versions,[ it looks up object using just the bucket name and object name]( https://github.com/fsouza/fake-gcs-server/blob/main/fakestorage/object.go#L517) and doesn't use generation id so it [defaults to using generation '0'](https://github.com/fsouza/fake-gcs-server/blob/665c3a68049611ce8d0006faaf94c4414fe87396/internal/backend/memory.go#L227). When generation 0 is used, it always does a lookup only in active objects so during first call it is able to find the object in activeObjects list but when it tries to find the object with different generation the next time since the generation is still set to 0, it errors out saying object not found.

AndrewKileC2FO (commented):
Is there any concern with the order the generations/versions are deleted?

funkyshu (changes requested):

seeroush (approved):

options/options.go
```diff
@@ -0,0 +1,5 @@
+package options
+
+type DeleteOption interface {
```
seeroush, line 3: Go-doc with examples on how to implement custom delete options might be handy here.

options/deleteOptions/deleteAllVersions.go
```diff
@@ -0,0 +1,15 @@
+package deleteOptions
+
+import "github.com/c2fo/vfs/v6/options"
+
+const OptionNameDeleteAllVersions = "deleteAllVersions"
+
+func WithDeleteAllVersion() options.DeleteOption {
```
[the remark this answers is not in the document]
dhondgepooja: Fixed the method name  as part of https://github.com/C2FO/vfs/pull/119/commits/814446f530b487b70276782a776edd38f337ee23

options/deleteOptions/deleteAllVersions.go
```diff
@@ -0,0 +1,15 @@
+package deleteOptions
```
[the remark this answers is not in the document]
dhondgepooja: Fixed the package name as part of this https://github.com/C2FO/vfs/pull/119/commits/814446f530b487b70276782a776edd38f337ee23

options/options.go
dhondgepooja: Added go doc https://github.com/C2FO/vfs/pull/119/commits/6bde3b81feda5219229324dd9c0ccc3a06a66a30#diff-76f121a3a81c8945dc0e0c81a4b95e941d252522eba8704a1aa6f6a1722a8b68R3-R13

dhondgepooja:
> You'll need to add entries in the changelog and don't forget to add/update any markdown as well.

Added changelog https://github.com/C2FO/vfs/pull/119/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

Added documentation for options and delete packages: 
https://github.com/C2FO/vfs/pull/119/files#diff-2e046795c81fb2bcc2667705f88c6421fa988047f147ddb2333ec8da5f638b7b
https://github.com/C2FO/vfs/pull/119/files#diff-64fb09ea01e76d4736694d316893f1ddaeb64208ca1b82838c2b01112220667b

Updated GS and S3 docs:
https://github.com/C2FO/vfs/pull/119/files#diff-87c3714d7e579142350666afd456d922a7dd3e7d55e174799dfa2bced366df57
https://github.com/C2FO/vfs/pull/119/files#diff-a52680507e0d3f24ae27c6e7e315ee16156cc1d32b0a16a0dafca33499cf6f30

funkyshu (approved):

funkyshu (approved):

PR #125, opened 2022-07-15
Update README.md

safaci2000:
Minor tweak, but go install doesn't take a -u flag.


funkyshu:
Thanks for noticing :)

PR #136, opened 2022-12-22
ftp backend

funkyshu:
### Added
- ftp backend to support connection via FTP, FTPS (implicit TLS), and FTPES (explicit TLS).


dmcilvain44 (approved):

funkyshu:
You might want to hold off pr'ing this for now.

I'm refactoring now because I found problems with the underlying lib, `github.com/jlaffaye/ftp`:
1.  There's a bug connecting to FTPeS that, so far, the maintainer has so far been reluctant to merge the fix for.  Either we'd have to wait for him or fork.
2. I ran into a problem when doing writes where the io.Pipe was getting closed from the read side before the write side was done.  I banged my head against this for a long time trying to fix it.
3. (there was another issue that I can't remember right now).

So now I'm switching to use `github.com/secsy/goftp`.  The downside here is I have to manually send the underlying FTP commands necessary to account for Seek/offsets.  Building in safety around this (checking for server-side support) isn't insurmountable but adds to the logic that was basically built-in with the other lib.  Finally refactor the rest of the rest of the code for the new lib interface.

dustin-schnee (approved):

funkyshu:
Closing.  Will open a new pr when I can get back to replacing the current ftp lib.

PR #147, opened 2023-11-21
Fix SFTP Backend to consider configured ports other than 22

dustin-schnee:
Currently the sftp filesystem implementation in vfs does not consider ports other than the standard 22, this change will make it so it leverages other ports if they are configured.


.github/workflows/whitesource.yml
dustin-schnee, line 1: This has been failing on all the recent PRs to this repo, we no longer use Whitesource, we use CodQL instead.

funkyshu (approved):

funkyshu (approved):

AnthonyRoss (approved):

PR #148, opened 2023-11-23
Stream on read 145

### LINKED ISSUE #145: Remove unnecessary write to temp when reading in S3 backend
**Is your feature request related to a problem? Please describe.**
Currently, in the S3 backend, anytime you call Read() or Seek(), the source file is copied locally.  This was likely originally done to accommodate easy use of os.Seek() when reading.  However, this means we can't truly stream from an `s3.File`.
This should improve copy time by removing the need to copy the contents locally.

**Describe the solution you'd like**
Remove the need for calling tempfile by tracking cursor positions when read, write, or seek  operations are called.  Use cursor position to set GetObjectInput.Range value to "seek" on Reads.

**Describe alternatives you've considered**
n/a

**Additional context**

funkyshu:
fixes #145 - Remove use of local temp file when reading/seeking from s3 files.  This should improve performance by allowing streaming reads from s3 files.

### DIFF
--- a/backend/s3/file.go
+++ b/backend/s3/file.go
@@ -341,14 +341,13 @@ func (f *File) Write(data []byte) (res int, err error) {
 		f.writeBuffer = bytes.NewBuffer([]byte{})
 	}
 
-	write, err := f.writeBuffer.Write(data)
+	written, err := f.writeBuffer.Write(data)
 	if err != nil {
 		return 0, err
 	}
-	offset := int64(write)
-	f.cursorPos += offset
+	f.cursorPos += int64(written)
 
-	return write, err
+	return written, err
 }
 
 // Touch creates a zero-length file on the vfs.File if no File exists.  Update File's last modified timestamp.
--- a/backend/s3/file.go
+++ b/backend/s3/file.go
@@ -283,7 +283,7 @@ func (f *File) Seek(offset int64, whence int) (int64, error) {
 	}
 
 	// update cursorPos
-	pos, err := seekTo(f.cursorPos, offset, whence, length, f.URI())
+	pos, err := seekTo(int64(length), f.cursorPos, offset, whence)
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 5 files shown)


backend/s3/file.go
```diff
@@ -534,8 +533,42 @@ func waitUntilFileExists(file vfs.File, retries int) error {
...
-	return s3manager.NewDownloaderWithClient(client, func(d *s3manager.Downloader) {
-		d.PartSize = partSize
-	})
+func (f *File) getReader() (io.ReadCloser, error) {
+	if f.reader == nil {
+		// Create the request to get the object
+		input := new(s3.GetObjectInput).
+			SetBucket(f.bucket).
+			SetKey(f.key).
+			SetRange(fmt.Sprintf("bytes=%d-", f.cursorPos))
```
funkyshu, line 542: This is the key change.  We use the cursor position as the startpoint for reading from getObject stream.  Anytime Seek or Close called, we invalidate the reader and any new call to Read opens a new stream at the updated cursor position.

backend/s3/file.go
```diff
@@ -534,8 +533,42 @@ func waitUntilFileExists(file vfs.File, retries int) error {
...
+	return f.reader, nil
+}
+
+func handleExistsError(err error) error {
+	if err != nil {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) {
+			switch awsErr.Code() {
+			case s3.ErrCodeNoSuchKey, s3.ErrCodeNoSuchBucket, "NotFound":
+				return vfs.ErrNotExist
```
funkyshu, line 568: This effectively converts aws errors to a vfs "file does not exist" error.

backend/s3/file.go
```diff
@@ -66,13 +56,10 @@ func (f *File) Path() string {
...
-		code = err.(awserr.Error).Code()
-	}
-	if err != nil && (code == s3.ErrCodeNoSuchKey || code == "NotFound") {
-		return false, nil
-	} else if err != nil {
+		if errors.Is(err, vfs.ErrNotExist) {
+			return false, nil
+		}
 		return false, err
 	}
```
funkyshu, line 65: this logic is largely moved to handleExistsError which is called directly in the getHeadObject call, so now we just look for the sentinel error, vfs.ErrNotExist.

backend/s3/file.go
```diff
@@ -104,8 +91,8 @@ func (f *File) Location() vfs.Location {
 // method if the target file is also on S3, otherwise uses io.CopyBuffer.
 func (f *File) CopyToFile(file vfs.File) error {
 	// validate seek is at 0,0 before doing copy
-	if err := backend.ValidateCopySeekPosition(f); err != nil {
-		return err
+	if f.cursorPos != 0 {
+		return vfs.CopyToNotPossible
 	}
```
funkyshu, line 96: We're now able to rely on the cursor position (since we track that now) to determine if it's safe to copy (cursor position must be zero)

dustin-schnee (approved):

PR #153, opened 2024-01-22
Fixed #152 bug where s3 backend failed to read empty files

### LINKED ISSUE #152: S3 backend errors when reading an empty file
**Describe the bug**
s3 backend returns an "InvalidRange: The requested range is not satisfiable" error when reading from an empty file.

**To Reproduce**
```go
func main() {
	_ = os.Setenv("AWS_REGION", "us-west-2")
	uri := "s3://c2foupload-test/empty.txt"

	// setup vfs s3 file to write to
	file, err := vfssimple.NewFile(uri)
	if err != nil {
		panic(err)
	}
	// create empty file
	ct, err := file.Write([]byte{})
	if err != nil {
		panic(err)
	}
	if ct != 0 {
		panic("expected 0 bytes written")
	}
	err = file.Close()
	if err != nil {
		panic(err)
	}

	// setup vfs s3 file to read from
	file, err = vfssimple.NewFile(uri)
	if err != nil {
		panic(err)
	}
	// confirm file is empty
	sz, err := file.Size()
	if err != nil {
		panic(err)
	}
	if sz != 0 {
		panic("expected 0 bytes written")
	}

	// read from file
	readCount, err := io.Copy(os.Stdout, file)
	if err != nil {
		panic(err) // it will panic here with: InvalidRange: The requested range is not satisfiable
	}
	if readCount != 0 {
		panic("expected 0 bytes read")
	}
}
```

**Expected behavior**
it should return with no error and zero bytes copied.

funkyshu:
S3 backend errors when reading an empty file #152

### DIFF
--- a/backend/s3/file.go
+++ b/backend/s3/file.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/url"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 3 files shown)


backend/s3/file.go
```diff
@@ -535,26 +536,35 @@ func waitUntilFileExists(file vfs.File, retries int) error {
 
 func (f *File) getReader() (io.ReadCloser, error) {
 	if f.reader == nil {
```
funkyshu, line 538: GH does a poor job of showing that I just moved this down to be nested under 'else'

dustin-schnee (approved):

dhondgepooja (approved):

PR #155, opened 2024-01-30
Sftp writes should truncate 154

funkyshu:


backend/sftp/file.go
```diff
@@ -260,11 +265,13 @@ func (f *File) Read(p []byte) (n int, err error) {
 	f.fileSystem.connTimerStop()
 	defer f.fileSystem.connTimerStart()
 
-	sftpfile, err := f.openFile(os.O_RDONLY)
+	sftpfile, err := f.openFile(os.O_RDWR)
```
funkyshu, line 268: This allows for (contrived but possible) sequence of reading some bytes in a file (for instance the first line (like csv header), then overwriting things after that.   Readonly would've discounted the possiblity or writing after read.

backend/testsuite/backend_integration_test.go
```diff
@@ -914,7 +861,7 @@ func (s *vfsTestSuite) File(baseLoc vfs.Location) {
 // gs-specific test cases
 func (s *vfsTestSuite) gsList(baseLoc vfs.Location) {
 	/*
-			test scenario:
+			test description:
```
funkyshu, line 864: Only moved contents of this file to common.go.

backend/testsuite/io_integration_test.go
```diff
@@ -0,0 +1,557 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+
+	"github.com/c2fo/vfs/v6"
+	"github.com/c2fo/vfs/v6/options"
+	"github.com/c2fo/vfs/v6/vfssimple"
+)
+
+type OSWrapper struct {
```
funkyshu, line 21: OSWrapper implements ReadWriteSeekCloseURINamer used by this test to generically test io functionality for vfs.File implemenations.  However, I needed a baseline to model the tests on so this basically provides an unix os file interface (outside of the os backend) to ensure we're following those io behaviors.

backend/testsuite/io_integration_test.go
```diff
@@ -0,0 +1,557 @@
...
+			"new text",
+		},
+		{
+			"Write, Seek, Read, Close, file exists",
+			"W(new text);S(0,0);R(5);C()",
+			true,
+			false,
+			"new text",
+		},
+	}
```
funkyshu, line 349: These tests cases use a mini-DSL to perform read/write/seek/close functions in various sequences.  It was surprising how little the SFTP implementation had a to change to match the tests.  Basically, only changes to a couple of flags and (making Read() and Write() use RW (vs read-only or write-only), and the conditional Truncate.

seeroush: Nice! I like the repeatable nature of this.

backend/testsuite/common.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/c2fo/vfs/v6/backend/azure"
+	"github.com/c2fo/vfs/v6/backend/ftp"
+	"github.com/c2fo/vfs/v6/backend/gs"
+	"github.com/c2fo/vfs/v6/backend/mem"
+	_os "github.com/c2fo/vfs/v6/backend/os"
+	"github.com/c2fo/vfs/v6/backend/s3"
+	"github.com/c2fo/vfs/v6/backend/sftp"
+)
+
+func CopyOsLocation(loc vfs.Location) vfs.Location {
```
seeroush, line 16: Doesn't look like our PR checks prevent merging on lint checks, but these are are missing the comments.

funkyshu: I think because they're in tests.  golangci is a collection of many linters.  I'll bet the one that deals with comments (golint/revive) skips for tests.

backend/testsuite/io_integration_test.go
seeroush: I might be missing some context around how these get executed. Is it expected that contributors add test cases here, or do we expect contributors to run this test suite specifically? Our README doc doesn't really call out how to perform integration tests.

backend/testsuite/common.go
funkyshu: I think because they're in tests.  golangci is a collection of many linters.  I'll bet the one that deals with comments (golint/revive) skips for tests.
It passes golangci-lint locally and in CI.  I did have to fix some other lint issues earlier. 
https://github.com/C2FO/vfs/blob/master/.golangci.yml#L215

backend/testsuite/io_integration_test.go
funkyshu: Yeah, I actually built these tests as part of a different PR for S3 streaming (vs saving to memory) but needed to "know" how to handle various sequences first.  My intent is to add each backend to the CI (with gh secrets).  Locally, you can run it the same way you do with the earlier (spaghetti-like) backend integration tests.
See https://github.com/C2FO/vfs/blob/master/backend/testsuite/doc.go

AnthonyRoss (approved):

dustin-schnee (approved):

PR #159, opened 2024-02-02
Updated sftp backend to to fix issue where some servers return a gene…

### LINKED ISSUE #158: Opening RW on read causes issues with some SFTP servers.
**Describe the bug**
A recent update to change Read() calls to open a file with os.O_RDWR returns a generic SSH_FX_Failure error on some servers.

Instead of of opening RW, we should open RO on initial Read and reopen in RW is Write is later called.  This is a bit of a contrived scenario (writing after reading) but it is possible.

funkyshu:
…ric error message when a file is opened for RW on Read(). Fixes #158.

### DIFF
--- a/backend/sftp/file.go
+++ b/backend/sftp/file.go
@@ -20,6 +20,7 @@ type File struct {
 	opener     fileOpener
 	seekCalled bool
 	readCalled bool
+	flagsUsed  int
 }
 
 // this type allow for injecting a mock fileOpener function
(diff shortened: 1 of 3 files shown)


backend/sftp/file.go
```diff
@@ -281,7 +282,7 @@ func (f *File) Seek(offset int64, whence int) (int64, error) {
 	f.fileSystem.connTimerStop()
 	defer f.fileSystem.connTimerStart()
 
-	sftpfile, err := f.openFile(os.O_RDWR)
+	sftpfile, err := f.openFile(os.O_RDONLY)
```
dustin-schnee, line 285: I should have asked this last time, but would you explain what calling openFile 2 times accomplishes?

dustin-schnee, line 285: I should have asked this last time, but would you explain what calling openFile 2 times accomplishes.

funkyshu: At this line? What do you mean?

dustin-schnee: Nevermind, ignore me, I failed to look at it closely :)

dustin-schnee (approved):

Issue #167, opened 2024-02-22
VFS should close source and target files during CopyToFile even if it errors


AnthonyRoss:
CopyToFile currently doesn’t close the source and target files in s3 implementation if there is an error during copy. However, it should always close both the source and target files whether there is an error or not during the copy.

https://github.com/C2FO/vfs/blob/master/backend/s3/file.go#L110C16-L110C26 

We should look into other implementations as well.

Ideally all implementations of CopyToFile should always close the source and target files if there is an error or not during copy

PR #168, opened 2024-02-22
Added defer to file.Close()

AnthonyRoss:
Resolves https://github.com/C2FO/vfs/issues/167

### DIFF
--- a/backend/azure/file.go
+++ b/backend/azure/file.go
@@ -140,7 +140,10 @@ func (f *File) CopyToLocation(location vfs.Location) (vfs.File, error) {
 // CopyToFile puts the contents of the receiver (f *File) into the passed vfs.File parameter.
 func (f *File) CopyToFile(file vfs.File) error {
 	// Close file (f) reader regardless of an error
-	defer f.Close()
+	defer func() {
+		_ = f.Close()
+		_ = file.Close()
+	}()
 
 	// validate seek is at 0,0 before doing copy
 	if err := backend.ValidateCopySeekPosition(f); err != nil {
--- a/backend/azure/file.go
+++ b/backend/azure/file.go
@@ -138,16 +138,26 @@ func (f *File) CopyToLocation(location vfs.Location) (vfs.File, error) {
 }
 
 // CopyToFile puts the contents of the receiver (f *File) into the passed vfs.File parameter.
-func (f *File) CopyToFile(file vfs.File) error {
+func (f *File) CopyToFile(file vfs.File) (err error) {
 	// Close file (f) reader regardless of an error
 	defer func() {
-		_ = f.Close()
-		_ = file.Close()
+		// close writer
+		wErr := file.Close()
+		// close reader
+		rErr := f.Close()
+		//
+		if err == nil {
+			if wErr != nil {
+				err = wErr
+			} else if rErr != nil {
+				err = rErr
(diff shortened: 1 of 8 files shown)


backend/s3/file.go
```diff
@@ -108,6 +108,8 @@ func (f *File) Location() vfs.Location {
 // CopyToFile puts the contents of File into the targetFile passed. Uses the S3 CopyObject
 // method if the target file is also on S3, otherwise uses io.CopyBuffer.
 func (f *File) CopyToFile(file vfs.File) error {
+	// Close file (f) reader regardless of an error
+	defer f.Close()
```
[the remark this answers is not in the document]
funkyshu: Yes, i'd say so.  Writes won't commit until you Close.  A subsequent Close is a noop so there's no danger if the the CopyToFIle caller closes.

backend/azure/file.go
```diff
@@ -139,6 +139,9 @@ func (f *File) CopyToLocation(location vfs.Location) (vfs.File, error) {
 
 // CopyToFile puts the contents of the receiver (f *File) into the passed vfs.File parameter.
 func (f *File) CopyToFile(file vfs.File) error {
+	// Close file (f) reader regardless of an error
+	defer f.Close()
```
funkyshu, line 143: `defer func() { _ = f.Close() }()`

AnthonyRoss: @funkyshu I thought about this but is there any reason that we would want to return the error from the Close() method?

backend/s3/file.go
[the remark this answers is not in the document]
AnthonyRoss: @dhondgepooja and @funkyshu fixed here: https://github.com/C2FO/vfs/pull/168/commits/012254fcc7bd86855f9055de11c2d98aee4cb35c

backend/azure/file.go
funkyshu: defer can't return an error, directly.  Close (on defer) is the most ignored error in go.  The only way you can really return an error in defer is by using a named error, but then you have protect against over writing a more important error.  Like lets say you're reading from a file and file doesn't exist.  That's the important thing to know.  Not that you couldn't also close the file, in defer.
So you have to do something like this: which may or may not be worth the time.  Having said that, the most important func in VFS is Close() because that's the single place where data is committed to the fileystem (on writes). So it MAY be worth ti.
```go
func (f *File) CopyToFile(file vfs.File) (err error) {
defer func() { 
    closeErr := f.Close()
    // only capture close err if there was no other error
    if closeErr != nil && err == nil {
          err = closeErr
    }
}()
```

An alternative is to concatenate multiple errors. 
```go
func (f *File) CopyToFile(file vfs.File) (err error) {
defer func() { 
    closeErr := f.Close()
    // only capture close err if there was no other error
    if closeErr != nil {
        if err == nil {
              err = closeErr
        } else {
              err = fmt.Errorf("orig error: %w - close err: %w", err, closeErr)
        } 
    }
}()

AnthonyRoss: Spoke with @funkyshu over slack and we cam eup with https://github.com/C2FO/vfs/pull/168/commits/1c2cce19f35944ce5cfaabe14dcddca3241905d3

funkyshu (approved):

PR #181, opened 2024-04-24
fixes #180 - SFTP backend now only opens files for W (instead of RW) …

### LINKED ISSUE #180: Only open sftp file as RW when Write'ing after a read action (Read or Seek)
**Describe the bug**
When doing only Write actions in SFTP backend, we unnecessarily open files as `os.O_RDWR`.  This causes issues if the location you're connecting to provides only write access, no read.

**Expected behavior**
The only time we need RW is when we write after a  read action (Read or Seek).  Otherwise a Write should open files as `os.O_WRONLY`.

**Additional context**
Currently, a Write to a directly that has insufficient permissions returns "write error: file not found".

funkyshu:
…except after read/seek. This should fix issues where server permission are write only

### DIFF
--- a/backend/sftp/file.go
+++ b/backend/sftp/file.go
@@ -24,7 +24,7 @@ type File struct {
 	flagsUsed  int
 }
 
-// this type allow for injecting a mock fileOpener function
+// this type allows for injecting a mock fileOpener function
 type fileOpener func(c Client, p string, f int) (ReadWriteSeekCloser, error)
 
 // Info Functions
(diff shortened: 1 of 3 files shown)


dustin-schnee (approved):

AnthonyRoss (approved):

backend/sftp/file.go
```diff
@@ -379,12 +379,41 @@ func (f *File) String() string {
 // openFile wrapper allows us to inject a file opener (for mocking) vs the defaultOpenFile.
 func (f *File) openFile(flags int) (ReadWriteSeekCloser, error) {
 	if f.sftpfile != nil {
-		FlagsUsedContainsReadOnly := f.flagsUsed&(os.O_WRONLY|os.O_RDWR) == 0
-		flagsContainsReadWrite := flags&os.O_RDWR != 0
+
+		// this case shouldn't normally exist except when we've set our own ReadWriteSeekCloser in tests
```
funkyshu, line 383: This was only an issue because the read-only flag `os.O_RDONLY` is equal to the zero-value of the struct member `flagsUsed`.  It is only relevant in tests where we've set our own ReadWriteSeekCloser but have never called a read action.

PR #183, opened 2024-05-15
Location exists bug 182

funkyshu:


backend/ftp/dataconn.go
```diff
@@ -152,7 +147,7 @@ func getDataConn(ctx context.Context, authority utils.Authority, fs *FileSystem,
 		fs.dataconn = nil
 	}
 
-	if fs.dataconn == nil || (f != nil && f.resetConn) {
+	if fs.dataconn == nil || fs.resetConn {
```
funkyshu, line 150: moved this reset flag from `f.restConn` to the `f.filesystem.resetConn` for the following scenario:
```go
_, _ = f.Write("something")
_ = f.Close() // this closes all dataconn connections and marks it as resetConn=true

otherFile := f.Location().NewFile("newfile.txt")
otherFile.Read(&buffer)
```
This would fail because Read would call `otherfile.filesystem.dataconn.R.Read()` but R is nil (thus the need for resetConn) but Location had no knowledege of the resetConn value.  So it really belongs at the filesytem level.

backend/ftp/dataconn_test.go
```diff
@@ -19,145 +19,184 @@ import (
 
 type dataConnSuite struct {
 	suite.Suite
+	ftpFile *File
+	client  *mocks.Client
```
funkyshu, line 23: most changes in this test are simple refactoring of 1 long, single test broken into multiple tests.  Other than that, updating the `List` asserts to match the parent directory "/", rather than the location dir "/some/" and returning  List entries that includes "some" as a directory entry. The refactor makes it look worse than it is.

backend/ftp/location.go
```diff
@@ -152,15 +156,13 @@ func (l *Location) Exists() (bool, error) {
...
-		return false, err
+	for _, entry := range entries {
+		if entry.Name == path.Base(l.Path()) && entries[0].Type == _ftp.EntryTypeFolder {
+			return true, nil
+		}
 	}
 
-	return true, nil
+	return false, nil
 }
```
funkyshu, line 166: This file is the actual fix.

dustin-schnee (approved):

backend/ftp/location.go
```diff
@@ -143,7 +143,11 @@ func (l *Location) Exists() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	entries, err := dc.List(l.Path())
+
+	// get parent directory by removing the last part of the path
+	paranetDir := strings.TrimSuffix(l.Path(), path.Base(l.Path())+"/")
```
grant-higgins-0, line 148: Should be parentDir

funkyshu: you don't like paranet? lol

PR #186, opened 2024-05-28
Fixed #185 - location.Exists was checking if a list entry was a direc…

### LINKED ISSUE #185: FTP backend Exists() checks if wrong entry is directory
**Describe the bug**
Calling Exists sometimes attempts to create a directory that already exists because we're checking for `entries[0]` rather the the current entry at https://github.com/C2FO/vfs/blob/v6.14.0/backend/ftp/location.go#L160

**To Reproduce**
Create a file and subdirectory in some base dir, then call Exists on the subdirectory.  If the file entry is returned first the entry type check will always be "file".

**Expected behavior**
Exists should properly check for the entry type of the current entry.

funkyshu:
…tory but it was was only checking the first entry.

### DIFF
--- a/backend/ftp/location.go
+++ b/backend/ftp/location.go
@@ -144,8 +144,9 @@ func (l *Location) Exists() (bool, error) {
 		return false, err
 	}
 
+	locBasename := path.Base(l.Path())
 	// get parent directory by removing the last part of the path
-	parentDir := strings.TrimSuffix(l.Path(), path.Base(l.Path())+"/")
+	parentDir := strings.TrimSuffix(l.Path(), locBasename+"/")
 
 	entries, err := dc.List(parentDir)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


katiehrenchir (approved):

backend/ftp/location.go
```diff
@@ -156,8 +157,8 @@ func (l *Location) Exists() (bool, error) {
 		return false, err
 	}
 
-	for _, entry := range entries {
-		if entry.Name == path.Base(l.Path()) && entries[0].Type == _ftp.EntryTypeFolder {
+	for i := range entries {
+		if entries[i].Name == locBasename && entries[i].Type == _ftp.EntryTypeFolder {
```
funkyshu, line 161: entries[0] was the culprit here... only checking the first entry's type, resulting Exists returning false, then MoveToFile tried to create the dir (which already existed).

PR #190, opened 2024-06-10
Fixed #189 - Update utils authority package to handle proper encoding…

### LINKED ISSUE #189: utils/authority doesn't properly handle userinfo percent-encoding
**Describe the bug**
Certain characters like `\` are disallowed in raw form in the userinfo section of a URI.  They must be percent/url -encoded.  However the vfs utility for handing authority does not handle decoding percent/url -encoded values.

**To Reproduce**
```go
uri := `sftp://domain.com%5cuser@host.com:22/`
loc, err := vfssimple.NewLocation(uri)
if err != nil {
	panic(err)
}
fmt.Println(loc.URI())
fmt.Println(loc.(*sftp.Location).Authority.UserInfo().Username())
```

**Expected behavior**
loc.URI() should returns the still-encoded uri as expected but the fetch the username should return the decoded value: `domain.com\user` but it returns `domain.com%5Cuser`.  This will pass the wrong username to the client which fails authentication.

**Additional context**
This will affect both FTP and SFTP.

funkyshu:
…/decoding of uri with reserved characters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
+## [6.14.3] - 2024-05-30
+### Fixed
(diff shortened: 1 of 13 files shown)


dustin-schnee (approved):
Just some lint cleanup to do.

dhondgepooja (approved):

PR #207, opened 2024-10-17
Windows support

NathanBaulch:
All unit tests now passing on Windows.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -27,7 +27,6 @@ jobs:
       - name: Set TMPDIR for Windows
         if: matrix.os == 'windows-latest'
         run: |
-          mkdir C:\Temp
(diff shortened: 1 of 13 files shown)


funkyshu (changes requested):
This looks good to me. Just add a CHANGELOG.md entry.
Also can you add `windows-latest` to the test matrix at https://github.com/C2FO/vfs/blob/main/.github/workflows/go.yml#L17

NathanBaulch:
Unit tests all pass on my Windows 10 machine - I guess we have more work to do to get it running in GitHub actions...

funkyshu (approved):

funkyshu:
@NathanBaulch This is a set of fixes I've meant to address for a long time but never had time for since we don't use Windows.  Thanks for this important fix.

PR #209, opened 2024-10-17
Fix typos

NathanBaulch:
Fixes a bunch of typos and minor grammar issues that I've stumbled on.

### DIFF
--- a/backend/sftp/options.go
+++ b/backend/sftp/options.go
@@ -27,7 +27,7 @@ type Options struct {
 	KnownHostsFile     string              `json:"knownHostsFile,omitempty"` // env var VFS_SFTP_KNOWN_HOSTS_FILE
 	KnownHostsString   string              `json:"knownHostsString,omitempty"`
 	KeyExchanges       []string            `json:"keyExchanges,omitempty"`
-	Ciphers            []string            `json:"cihers,omitempty"`
+	Ciphers            []string            `json:"ciphers,omitempty"`
 	MACs               []string            `json:"macs,omitempty"`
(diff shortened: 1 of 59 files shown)


backend/sftp/options.go
```diff
@@ -27,7 +27,7 @@ type Options struct {
 	KnownHostsFile     string              `json:"knownHostsFile,omitempty"` // env var VFS_SFTP_KNOWN_HOSTS_FILE
 	KnownHostsString   string              `json:"knownHostsString,omitempty"`
 	KeyExchanges       []string            `json:"keyExchanges,omitempty"`
-	Ciphers            []string            `json:"cihers,omitempty"`
+	Ciphers            []string            `json:"ciphers,omitempty"`
```
funkyshu, line 30: oh wow... that's a good catch... this is probably breaking for someone, somewhere.

funkyshu (changes requested):
I don't know if you love doing tedious combing over docs/code or you have a wonderful spell checker you're doing "god's work" here.
The sheer volume of spelling and grammatical errors is egregious (and most of them are probably mine).
Seems like its time for a better spell checker or we need to tweak golangci's spelling linter(s).

Only one thing, like the test improvement PR, please add an entry to the CHANGELOG.md, otherwise, this is much appreciated work!!!!

backend/sftp/options.go
NathanBaulch: Can we support reading both fields but always write the correct spelling when marshaling moving forward?

NathanBaulch:
Heh, I use a combination of the [typos](https://github.com/crate-ci/typos) tool along with the JetBrains IDE language inspections.

funkyshu (approved):

PR #211, opened 2024-10-18
Configure mockery interfaces

NathanBaulch:
Noticed that some mocks weren't being regenerated because they weren't included in the mockery config file.


funkyshu (approved):

funkyshu (changes requested):
Add CHANGELOG.md entry (might rebase first to catch your other entries now merged to main).

funkyshu (approved):

PR #255, opened 2025-04-21
add contrib/lockfile

### LINKED ISSUE #254: Feature Request: Implement File Locking Mechanism
# Feature Request: Implement File Locking Mechanism

**Problem**

Currently, when multiple processes or application instances interact concurrently with the same file resource via `vfs` (especially on shared backends like OS, SFTP, or potentially cloud storage depending on usage patterns), there's no standard mechanism provided by the `vfs` interfaces to prevent race conditions or ensure exclusive access during critical operations (e.g., read-modify-write cycles). This can lead to data corruption or unpredictable application behavior.

**Proposal**

I propose adding a file locking mechanism to the `vfs.File` interface. This could take the form of methods like:

* `Lock() error`: Acquires an exclusive lock (blocking).
* `TryLock() (bool, error)`: Attempts to acquire an exclusive lock (non-blocking).
* `RLock() error`: Acquires a shared/read lock (blocking).
* `TryRLock() (bool, error)`: Attempts to acquire a shared/read lock (non-blocking).
* `Unlock() error`: Releases the lock held by the file handle.

**Considerations**

1.  **Backend Support**: File locking semantics and capabilities vary significantly across backends (OS, SFTP, S3, GCS, Azure, FTP, mem).
    * How should backends that don't natively support locking handle these methods? (e.g., return a specific `ErrNotSupported`?)
    * OS/SFTP might map reasonably well to `flock` or `fcntl`.
    * Cloud storage might require different strategies (e.g., using leases/metadata, or potentially being unsupported for standard locking).
    * The `mem` backend could implement simple mutex-based locking.
2.  **Lock Type**: Primarily advisory locks seem most feasible across different systems. Mandatory locking is often OS-specific and harder to abstract.
3.  **Scope**: Locks would typically be associated with the `vfs.File` handle/instance.

**Question**

Is there any existing functionality or recommended pattern within `c2fo/vfs` to handle file locking or achieve exclusive access currently? I couldn't find explicit mentions of a locking API in the documentation or recent issues.

Implementing such a feature, even if only for a subset of backends initially, would significantly enhance the robustness of applications using `c2fo/vfs` in concurrent environments.

Thanks for considering!

funkyshu:
Fixes #254

### DIFF
--- a/contrib/lockfile/lockfile.go
+++ b/contrib/lockfile/lockfile.go
@@ -1,0 +1,254 @@
+/*
+Package lockfile provides an advisory locking mechanism for vfs.File objects
+using companion `.lock` files.
+
+⚠️ Advisory Locking Note:
+
+This package implements **advisory locks**, which means:
+- Locks are **not enforced by the OS, filesystem, or remote backend**
+- All cooperating processes must **explicitly check for and honor** the lock
+- It is still possible for non-cooperative processes to ignore the lock and access the file
+
+Lock files are created using atomic file creation (CreateExclusive + MoveTo)
+and include metadata such as timestamp, PID, hostname, and optional TTL.
+
+This approach is portable and backend-agnostic, and it works on local filesystems,
+cloud object storage (e.g., S3/GCS), SFTP, and any vfs backend that supports
+basic file creation and deletion.
+
+This is similar in spirit to common conventions used by package managers, editors,
+and distributed systems where mandatory locking is unavailable or unreliable.
+*/
+package lockfile
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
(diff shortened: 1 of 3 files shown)


contrib/lockfile/lockfile.go
```diff
@@ -0,0 +1,254 @@
...
+
+	"github.com/c2fo/vfs/v7"
+)
+
+var (
+	ErrLockAlreadyHeld = errors.New("lockfile: lock already held")
+	ErrLockNotHeld     = errors.New("lockfile: no lock held")
+)
+
+type Metadata struct {
```
dustin-schnee, line 39: Just some missing comments in here.

dustin-schnee (approved):

PR #280, opened 2025-08-29
More consistent unit test assertions

NathanBaulch:
Fixes a variety of inconsistencies in unit test assertions.
* Avoid `True()` if a more specific assertion exists
* Use `False()` instead of `True(!...)`
* Use `ErrorContains()` instead of `Contains(err.Error(),...)`
* Use `Positive()` instead of `Greater(...,0)`
* Use `Empty()` instead of `Len(...,0)`
* Use `Zero()` instead of `Equal(0,...)`
* Use `Equal()` instead of `EqualValues()`
* Use embedded assert methods on `suite.Suite` directly
* Use `IsType()` when checking types
* Use `Len()` when checking slice or string length
* Ensure expected and actual arguments passed to `Equal` are in the correct order


NathanBaulch:
Note that a lot of these issues were identified by enabling `testifylint` in the golangci-lint config. Probably a good idea to permanently enable this.

funkyshu (approved):

funkyshu:
were these also found with testify lint?

NathanBaulch:
Some of them were. The next step was to enable that in the project now that it's probably close to passing.

PR #284, opened 2025-09-15
Lint vfsintegration test code

NathanBaulch:
Noticed that the `vfsintegration` test code wasn't being checked by golangci-lint.


funkyshu:
@NathanBaulch 
Thanks for all your submissions.  Love the additional tidiness!!

PR #285, opened 2025-09-15
Use utils.Ptr where possible

NathanBaulch:
The shared `utils.Ptr` func should be used everywhere.


NathanBaulch:
Flakey test maybe?

funkyshu (approved):

PR #286, opened 2025-09-16
Additional linting

NathanBaulch:
This PR enables 9 more golangci-lint linters that I personally use, along with compliance fixes.
I've kept each in a separate commit so they can be selectively removed with ease if desired. I'm fine with squashing the PR on merge, or splitting one or more into separate PRs too.
The only potentially disruptive one is `tagliatelle` since it renames JSON fields.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,11 @@
 GOBIN ?= $$(go env GOPATH)/bin
+MODULES := $(shell find . -type f -name "go.mod" -exec dirname {} \;)
 
 .PHONY: lint
-lint:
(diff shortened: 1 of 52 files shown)


Makefile
```diff
@@ -1,8 +1,11 @@
 GOBIN ?= $$(go env GOPATH)/bin
+MODULES := $(shell find . -type f -name "go.mod" -exec dirname {} \;)
```
funkyshu, line 2: 👍

funkyshu (approved):

PR #307, opened 2025-12-19
Aws role pass in credentials

katiehrenchir:
Addresses https://github.com/C2FO/vfs/issues/308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
-- AWS S3 role based auth for non-default profiles
+- AWS S3 role based auth for non-default profiles. Fixes #308.
 - Inline test case structs for consistency.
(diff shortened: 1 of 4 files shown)


backend/s3/options_test.go
```diff
@@ -81,6 +81,21 @@ func (o *optionsTestSuite) TestGetClient() {
...
+				Region:          "some-region",
+				RoleARN:         "arn:aws:iam::123456789012:role/my-role",
+			},
+			expected: func(o *optionsTestSuite, client *s3.Client, err error) {
+				o.Require().NoError(err)
+				o.NotNil(client, "client is set")
+				o.Equal("some-region", client.Options().Region, "region is set")
+				o.NotNil(client.Options().Credentials, "credentials are set")
+			},
+		},
```
funkyshu: These tests (the preexisting ones this is modeled after) really aren't great, are they? I mean we just know that the there are credentials but not necessarily that they are correct.  🤷‍♂️

funkyshu:
I added some doc updates.

dhondgepooja (approved):

PR #342, opened 2026-07-16
fix(vfsevents): fix CI-observed flake in s3events TestEnhancedMetadata

funkyshu:
## Summary

Fixes a CI failure observed on ubuntu + go1.26 after #339 merged:

```
--- FAIL: TestS3WatcherTestSuite/TestEnhancedMetadata
    SqsClient.go:23: FAIL: DeleteMessage(string,string)
    SqsClient.go:23: FAIL: 1 out of 2 expectation(s) were met.
        The code you are testing needs to make 1 more call(s).
```

## Root cause

`TestEnhancedMetadata` only synchronized on the event handler firing (via an `eventReceived` channel), not on `pollOnce`'s own completion. This meant the subtest could reach its assertions and finish (triggering mock-expectation teardown) before the background `pollOnce` goroutine made its `DeleteMessage` call — a timing gap that's tight enough to usually pass locally but wide enough to occasionally lose the race on CI runners.

This is exactly the residual race risk flagged (but intentionally not fixed at the time, since it wasn't the primary target) during review of #339's fix for the sibling `TestNonVersionedBucketMetadata` test.

## Fix
- `TestEnhancedMetadata` now uses the same completion-based synchronization as `TestNonVersionedBucketMetadata` — waits for `pollOnce` to fully return via an error channel, rather than only waiting for the handler to fire.
- Extracted the shared wait/cancel/grace-period logic (previously only in `TestNonVersionedBucketMetadata`) into a `waitForPoll` suite helper, now used by both tests, mirroring the equivalent `waitForReceive` helper added to `gcsevents_test.go` in #341.
- Uses `time.NewTimer` + `defer timer.Stop()` throughout (rather than `time.After`), proactively addressing the un-stopped-timer issue raised in review of #341.
- Updated `contrib/vfsevents/CHANGELOG.md`.

## Test plan
- `go test -race -run TestS3WatcherTestSuite -count=300 ./contrib/vfsevents/watchers/s3events/...` — clean.
- `go test -run TestS3WatcherTestSuite -count=1000 ./contrib/vfsevents/watchers/s3events/...` (no `-race`, to stress scheduling/timing rather than just data races) — clean.
- `go build ./... && go vet ./...` for `contrib/vfsevents` — clean.
- `golangci-lint run ./watchers/s3events/...` — only a pre-existing, unrelated `nolintlint` finding in `s3events.go` (not touched by this PR).

### DIFF
--- a/contrib/vfsevents/CHANGELOG.md
+++ b/contrib/vfsevents/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- `TestS3WatcherTestSuite/TestEnhancedMetadata`: fixed a CI-observed flake where the test could finish (and its mock-expectation teardown could run) before the background `pollOnce` goroutine's `DeleteMessage` call happened, since the test only synchronized on the event handler firing rather than on `pollOnce` fully returning. Now uses the same completion-based synchronization as `TestNonVersionedBucketMetadata` (extracted into a shared `waitForPoll` helper), which also proactively avoids the un-stopped `time.After` timer issue raised in review of [#341](https://github.com/C2FO/vfs/pull/341).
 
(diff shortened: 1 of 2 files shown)


funkyshu:
Folding this into #341 per discussion — same underlying pattern, and consolidating avoids two overlapping open PRs against the same test files.

PR #345, opened 2026-07-17
fix(vfsevents): fix fsnotify Windows drive-letter path round-trip

### LINKED ISSUE #344: vfsevents fsnotify: file:// URI round-trip produces invalid path on Windows
## Summary

The \`fsnotify\` watcher's test helper \`fileURL()\` (in \`contrib/vfsevents/watchers/fsnotify/fsnotify_test.go\`) builds a canonical \`file:///C:/Temp/...\` URI from a Windows drive-letter path and passes it to \`vfssimple.NewLocation\`. When that location's path is later handed to the underlying \`fsnotify.Add()\` call (or similar), it retains a leading slash before the drive letter (e.g. \`/C:/Temp/...\`), which Windows rejects:

\`\`\`
failed to add watch path: failed to add watch path /C:/Temp/TestDebouncingEdgeCases3938112569/001: GetFileAttributes: The filename, directory name, or volume label syntax is incorrect.
\`\`\`

This was discovered while fixing #338 and enabling \`-race\` + a proper \`TEMP\`/\`TMP\` setup in CI (see PR #343) — it was previously masked because the Windows CI job's \`TEMP\`/\`TMP\` env vars were silently broken (set to \`\`\`\`\`\`), so \`t.TempDir()\` was returning paths that didn't exercise this code path in the same way.

## Where to look

- \`contrib/vfsevents/watchers/fsnotify/fsnotify_test.go\`'s \`fileURL()\` helper (test-only) builds the URI.
- The bug itself is likely in how a \`file://\` URI with a Windows drive letter gets converted back into a native path — either in \`vfssimple\`/the OS backend's location path parsing, or in the fsnotify watcher's production code that converts a \`vfs.Location\` into a native path for \`fsnotify.Add()\`.

## Current state

The affected tests are temporarily skipped on \`windows\` (via \`t.Skip\`) in PR #343:
- \`TestFSNotifyWatcherTestSuite\` (all subtests, via \`SetupTest\`)
- \`TestDebouncing\`
- \`TestDebouncingEdgeCases\`
- \`TestEventAnalysis\`

## Next steps

- Reproduce locally on Windows (or in a Windows CI job) with debug logging of the path at each conversion step.
- Fix the drive-letter path handling so these tests can be un-skipped.

funkyshu:
## fsnotify: Windows drive-letter `file://` URI round-trip produces an invalid path (#344)

The fsnotify watcher's `Start()` built the path passed to `fsnotify.Add()` by naively stripping the `"file://"` prefix off the location's URI (`file:///C:/Temp/...` -> `/C:/Temp/...`). For a Windows drive-letter path this leaves a stray leading slash, which Windows rejects:

```
failed to add watch path: failed to add watch path /C:/Temp/TestDebouncingEdgeCases3938112569/001: GetFileAttributes: The filename, directory name, or volume label syntax is incorrect.
```

`convertEvent()` had the mirror problem on the way back out: it built the event's URI by naively prepending `"file://"` to the native OS path fsnotify reports (`C:\Temp\...`), producing an invalid URI with backslashes and no leading slash before the drive letter.

### Fix

Added `vfsPathToNative`/`nativePathToURI` helpers (each with a GOOS-parameterized `...OS` variant so the Windows-specific behavior can be unit tested from any platform) that correctly convert between VFS-style paths (forward slashes, leading slash before a drive letter) and native OS paths in both directions. `Start()` now uses `vfsPathToNative` before calling `fsnotify.Add()`, and `convertEvent()` now uses `nativePathToURI` when building the event URI.

### Tests

- Re-enabled the `fsnotify` watcher tests that were temporarily skipped on Windows for this issue in #343 (`TestFSNotifyWatcherTestSuite`, `TestDebouncing`, `TestDebouncingEdgeCases`, `TestEventAnalysis`), updating a couple of assertions that previously assumed a native path could be substring/equality-compared against a `file://` URI.
- Added `path_test.go` with GOOS-parameterized unit tests for the new conversion helpers (drive-letter stripping/re-adding, separator conversion, round-trip) so the Windows code path is verified without needing a Windows runner.

Fixes #344


funkyshu:
Pushed a follow-up commit addressing self-review feedback:

- **Drive-root paths**: a location watching an entire Windows drive (e.g. `C:\`) was being reduced to a bare `"C:"`, which Windows interprets as "current directory on drive C", not the drive root. `vfsPathToNativeOS` now re-adds the trailing separator when stripping leaves a bare drive letter.
- **Authority-safety**: `Start()` now derives the watch path via `Location.Path()` instead of hand-parsing `Location.URI()`, so a location with a non-empty authority (e.g. `file://host/C:/Temp`) can no longer have the authority swept into the path and mangled by the drive-letter conversion. Extracted into a small, directly testable `resolveWatchPath`/`resolveWatchPathOS` pair.
- **Authority round-trip**: `convertEvent()`'s URI conversion now preserves the location's authority instead of silently dropping it.
- Cross-referenced the near-duplicate drive-letter logic in `backend/os/file.go`'s `toNativeOSPath` (can't import across module boundaries) so the two stay in sync going forward.
- Extended `path_test.go` with cases for the drive-root fix, authority preservation, and a new `resolveWatchPathOS` suite that builds real (virtual) `vfs.Location`s via `vfssimple` to exercise the full location -> native-path path, still without needing a Windows runner.

grant-higgins-0 (approved):

jkware (approved):

Issue #317, opened 2026-03-06
Migrate S3 backend from deprecated manager APIs to feature/s3/transfermanager


funkyshu:
**Is your feature request related to a problem? Please describe.**

The S3 backend currently uses deprecated AWS SDK v2 APIs (`manager.NewUploader` and `manager.NewDownloader`) for file upload and download operations. These APIs are marked as deprecated in favor of the new `feature/s3/transfermanager` package. This generates staticcheck warnings (SA1019) across the codebase:

- `backend/s3/file.go:336` - `manager.NewUploader` is deprecated
- `backend/s3/file.go:340` - `uploader.Upload` is deprecated  
- `backend/s3/file.go:641` - `manager.NewDownloader` is deprecated
- `backend/s3/file.go:809` - `manager.NewUploader` is deprecated
- `backend/s3/file.go:817` - `uploader.Upload` is deprecated
- `backend/s3/file.go:847-848` - `manager.Downloader` type is deprecated
- `backend/s3/file.go:853-854` - `manager.Uploader` type is deprecated

While these APIs still function correctly, they may be removed in future AWS SDK releases, and the new transfermanager provides improved performance and better resource management.

**Describe the solution you'd like**

Migrate the S3 backend to use the new `github.com/aws/aws-sdk-go-v2/feature/s3/manager` transfermanager APIs:

1. Replace `manager.NewUploader` with the new transfermanager upload APIs
2. Replace `manager.NewDownloader` with the new transfermanager download APIs
3. Update the `withUploadPartitionSize` and `withDownloadPartitionSize` helper functions to work with the new API
4. Ensure partition size configuration options continue to work as expected
5. Verify that all upload/download scenarios (temp file writes, streaming writes, copy operations) work correctly with the new APIs

The migration should maintain backward compatibility with existing Options and configuration while eliminating the deprecation warnings.

**Describe alternatives you've considered**

1. **Keep using deprecated APIs with nolint directives** (current approach) - This works but kicks the can down the road and may break when AWS SDK eventually removes these APIs
2. **Wait for AWS SDK to force the migration** - Not recommended as it could break production code unexpectedly
3. **Implement a compatibility layer** - Would add unnecessary complexity when direct migration is feasible

**Additional context**

- AWS SDK discussion on the deprecation: https://github.com/aws/aws-sdk-go-v2/discussions/3306
- The current implementation uses partition size configuration for performance tuning, which must be preserved in the migration
[...]

PR #351, opened 2026-08-25
feat(s3): use ListObjectsV2 for listing, adapting legacy clients

funkyshu:
## Summary

Moves the s3 backend's listing from the older `ListObjects` operation to `ListObjectsV2`, without changing the exported `Client` interface. This is groundwork for #317: `transfermanager.S3APIClient` requires `ListObjectsV2`, so this removes the one blocker that would otherwise have forced a breaking change on that migration.

- **Listing uses `ListObjectsV2`.** `fullLocationList` now pages with `ContinuationToken` instead of `Marker`, which also removes a nil dereference of `IsTruncated`.
- **`Client` is untouched.** An unexported `backendClient` describes what the backend uses internally. A supplied client that lacks `ListObjectsV2` is wrapped in `legacyClient`, which emulates it over v1 `ListObjects`. `Client`, `WithClient`, and `FileSystem.Client()` keep their exact signatures and behavior, and `mocks/Client.go` regenerated to a zero-byte diff. The AWS SDK's `*s3.Client` satisfies `backendClient` directly, so the default path is never wrapped.
- **`backendClient` spells out its methods** rather than embedding `Client`, which both keeps the superseded `ListObjects` out of reach of backend code and leaves the internal contract free of any dependency on the deprecated `feature/s3/manager` package. It is a strict superset of `transfermanager.S3APIClient`, so it will satisfy that interface as-is.

### Two quiet failures now fail loudly

- A page reporting `IsTruncated` with no continuation token previously reissued the identical request **forever**. It now returns an error. The loop structure predates this PR, but the adapter can produce that shape, so the guard belongs here.
- A client that doesn't implement `s3.Client` (or a nil client) was silently discarded in favor of a default client, so callers could unknowingly run against real AWS credentials. Both injection paths now report it. Neither can return an error directly, so it surfaces on the next `Client()` call. **Behavior change:** code that previously passed an unusable client and got a default one will now see an error. Filed under `Fixed` as the old behavior was a bug; happy to reclassify if reviewers disagree.

## Verification

- `go test ./...` and `go test -race ./...` pass; `golangci-lint run ./...` reports 0 issues.
- **The five pre-existing location tests pass untouched while now running through the adapter**, including the paged case that asserts an exact `Marker` on the second call. That is behavioral equivalence between v1 and the emulation, not tests written to match the implementation.
- **Validated against a real server:** the testcontainers integration suite (`-tags=vfsintegration -race`) passes against MinIO for both the SSE-on and SSE-off s3 registrations. The `Location` conformance subtest exercises `List`, `ListByPrefix`, and `ListByRegex`, so the migrated code path is covered end to end, not just against mocks.
- Two adapter assumptions were checked empirically against MinIO with a throwaway probe rather than assumed:
  - `KeyCount` counts common prefixes alongside keys (`KeyCount=5` for `Contents=2` + `CommonPrefixes=3`), so the adapter sums both. My first version undercounted.
  - v1 `ListObjects` populates `NextMarker` when a delimiter is set, even for a page containing only common prefixes — the case that could otherwise have stranded pagination.
- Adapter is at 100% statement coverage; package total 84.3%, above the 80% gate.

## Known gap

`legacyClient` itself has never run against a real server — MinIO takes the native path, so the integration suite doesn't reach it. The probe above validated the two S3 semantics it depends on, which covers most of that risk. Closing it fully would mean a suite test that hides `ListObjectsV2` from MinIO's client and pages with `MaxKeys=1`.

## Not in scope

The transfermanager migration itself, and the AWS SDK dependency bump it needs. #317 stays open; all nine `nolint:staticcheck // SA1019` directives in `file.go` are untouched. Per the sequencing agreed in that issue, the swap waits until `feature/s3/transfermanager` leaves `v0.x`.

Refs #317


backend/s3/client.go
```diff
@@ -17,3 +19,120 @@ type Client interface {
...
+	out, err := c.ListObjects(ctx, &s3.ListObjectsInput{
+		Bucket:                   in.Bucket,
+		Delimiter:                in.Delimiter,
+		EncodingType:             in.EncodingType,
+		ExpectedBucketOwner:      in.ExpectedBucketOwner,
+		Marker:                   marker,
+		MaxKeys:                  in.MaxKeys,
+		OptionalObjectAttributes: in.OptionalObjectAttributes,
+		Prefix:                   in.Prefix,
+		RequestPayer:             in.RequestPayer,
```
dustin-schnee, line 99: Will this panic if a client passes in (nil, nil)? Would a nil check on the ListObjectsInput prevent it?

backend/s3/location_test.go
```diff
@@ -99,6 +100,83 @@ func (lt *locationTestSuite) TestList_pagedCall() {
...
+	loc, err := (&FileSystem{client: client}).NewLocation(bucket, "/dir1/")
+	lt.Require().NoError(err)
+
+	fileList, err := loc.List()
+	lt.Require().ErrorIs(err, errTruncatedWithoutToken)
+	lt.Empty(fileList)
+}
+
+// TestList_listObjectsV2Error guards against a lost or altered error when ListObjectsV2 itself
+// fails, whether on the first page or a later one.
```
dustin-schnee, line 158: The comment suggests that it will test the case where there is a truncated first page and then an error on a later page, but it seems to only test error on the first page. Perhaps remove the `or a later one.` or expand the test to comply with the comment?

dustin-schnee (approved):

